PR #2, opened 2026-01-28
ERC20 Integration

micbakos:
* Simple scaffolding of ERC20 transfers and balance checks.
* Tried to keep the current structure of packages as is so as to not mess up with ongoing work.
* Implementation of a script that fetches the top held erc20s and stores them as presets. The script can change to run on sepolia to create presets for that network too. Updated to use AVNU verified tokens list.
* Made a first step to type-safe starknet Addresses. This was the bare minimum. I would like your feedback on that. The reason behind this change was that besides the benefits of always running sanitization on string addresses (we could also offer checksum checks), we could later support strk domain addresses (this would require some change to Address becoming an interface, but I would prefer we started early using a type rather than strings).
* I would like your opinion on amounts. As per the documentation I kept it as string for now and didn't perform any checks. Do you think we should do similar types for amounts and create factories for people wanting to express the amount in formatted numbers and not in WEI or FRI etc..

### DIFF
--- a/src/types/amount.ts
+++ b/src/types/amount.ts
@@ -1,20 +1,5 @@
+import type { BigNumberish } from "starknet";
 import type { Token } from "./token.js";
-
-/**
- * Input type for creating Amount instances.
- * Accepts string, number, or bigint values.
- *
- * @example
- * ```ts
- * // All of these are valid AmountInput values:
- * "1.5"      // string with decimals
- * "100"      // string integer
- * 1.5        // number with decimals
- * 100        // number integer
- * 100n       // bigint
- * ```
- */
-export type AmountInput = string | number | bigint;
 
 /**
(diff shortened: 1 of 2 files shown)


src/types/amount.ts
```diff
@@ -0,0 +1,756 @@
...
+ * ```ts
+ * // All of these are valid AmountInput values:
+ * "1.5"      // string with decimals
+ * "100"      // string integer
+ * 1.5        // number with decimals
+ * 100        // number integer
+ * 100n       // bigint
+ * ```
+ */
+export type AmountInput = string | number | bigint;
```
0xLucqs: isn't that just `BigNumberish`?

micbakos: You are absolutely right. I will use that instead.

PR #6, opened 2026-02-02
refactor: wallet architecture and starknet.js v9 compatibility

0xLucqs:
## Summary

- Add SignerAdapter for bridging SDK signers to starknet.js v9
- Refactor Wallet to use composition instead of inheritance
- Add shared wallet utilities (utils.ts)
- Simplify Wallet.create() API with options object
- Support loading wallet with known address (skips computation)
- Update ArgentPreset for v0.4.0 format with CairoCustomEnum
- Add ArgentXV050Preset for Privy compatibility
- Add Address type for compile-time safety
- Add .env to gitignore for security

## Test plan

- [x] All existing tests pass
- [x] TypeScript compiles without errors
- [x] ESLint passes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -236,6 +236,22 @@
         color: var(--text-primary);
       }
 
+      .btn-copy {
+        background: transparent;
+        border: 1px solid var(--border-color);
+        border-radius: 6px;
+        padding: 0.25rem 0.5rem;
+        cursor: pointer;
+        font-size: 0.8rem;
+        transition: all 0.2s;
+        margin-left: 0.5rem;
+      }
+
+      .btn-copy:hover {
+        background: var(--bg-tertiary);
+        border-color: var(--accent-primary);
+      }
+
       .address {
         max-width: 200px;
         overflow: hidden;
@@ -404,7 +420,7 @@
       <div id="connect-section" class="card">
         <div class="card-title">Connect Wallet</div>
 
-        <div class="connect-grid">
+        <div class="connect-grid" style="grid-template-columns: 1fr 1fr">
           <button id="btn-cartridge" class="btn btn-cartridge">
             <svg
               width="20"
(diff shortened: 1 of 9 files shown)


src/wallet/index.ts
```diff
@@ -5,23 +5,55 @@ import {
...
+  /** Account: either AccountProvider or { signer, accountClass? } */
+  account:
+    | AccountProvider
+    | { signer: SignerInterface; accountClass?: AccountClassConfig };
+  /** RPC provider */
+  provider: RpcProvider;
+  /** SDK configuration */
+  config: SDKConfig;
+  /** Known address (skips address computation if provided) */
+  address?: string;
```
micbakos: Does this imply a known account address? If so should we rename to `accountAddress`?

Also you could use Address type here and better not to use a string?

src/wallet/index.ts
```diff
@@ -34,237 +66,190 @@ export { type WalletInterface } from "./interface.js";
...
- * const wallet = await sdk.connectWallet({
- *   account: {
- *     signer: new StarkSigner(privateKey),
- *     accountClass: OpenZeppelinPreset,
- *   }
+ * const wallet = await Wallet.create({
+ *   signer: new StarkSigner(privateKey),
+ *   accountClass: ArgentPreset,
+ *   provider,
+ *   config,
```
micbakos, line 73: Much better

src/wallet/utils.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Shared wallet utilities.
+ * Used by wallet implementations to avoid code duplication.
+ */
+
+/**
+ * Check if an account is deployed on-chain.
+ */
+export async function checkDeployed(
+  provider: RpcProvider,
+  address: string
```
micbakos: I would prefer all inputs to the system to be type-checked. What do you think?

0xLucqs: indeed, got changed in the rebase, reverted to correct code

src/wallet/index.ts
0xLucqs: Fixed

PR #8, opened 2026-02-02
feat: add WebAuthn signer for Face ID / Touch ID authentication

0xLucqs:
## Summary

- Add WebAuthnSigner for passkey-based authentication
- Add WebAuthnPreset for P-256 account contracts
- Support parseP256PublicKey for coordinate extraction
- Enable biometric authentication on supported devices

WebAuthn enables passwordless authentication using device biometrics (Face ID, Touch ID, Windows Hello) or security keys.

## Test plan

- [x] All existing tests pass
- [x] TypeScript compiles without errors
- [x] ESLint passes
- [ ] Manual test on device with biometrics

**Depends on:** #7

**Note:** WebAuthnPreset.classHash is currently a placeholder (0x0). A P-256 account contract needs to be deployed on Starknet before this can be used in production.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/account/presets.ts
+++ b/src/account/presets.ts
@@ -6,6 +6,7 @@ import {
   type Calldata,
 } from "starknet";
 import type { AccountClassConfig } from "../types/wallet.js";
+import { parseP256PublicKey } from "../signer/webauthn.js";
 
 /**
  * Devnet account preset.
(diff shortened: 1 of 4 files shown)


src/account/presets.ts
```diff
@@ -78,3 +79,54 @@ export const ArgentXV050Preset: AccountClassConfig = {
...
+ *
+ * const wallet = await sdk.connectWallet({
+ *   account: {
+ *     signer,
+ *     accountClass: WebAuthnPreset,
+ *   },
+ * });
+ * ```
+ */
+export const WebAuthnPreset: AccountClassConfig = {
```
micbakos, line 107: Maybe we shouldn't export the const but rather export a function like
```
webAuthnPreset(classHash: string) 
```
which will return the preset?

src/signer/webauthn.ts
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    const rpId = options?.rpId ?? window.location.hostname;
+    const rpName = options?.rpName ?? (document.title || "App");
+    const userId = options?.userId ?? crypto.randomUUID();
+    const userName = options?.userName ?? "User";
+
+    const credential = (await navigator.credentials.create({
+      publicKey: {
+        challenge: crypto.getRandomValues(new Uint8Array(32)),
```
micbakos, line 75: I think this might need to be changed to use mobile's secure rng? 

Should we start manual testing the sdk on a mobile friendly environment?

0xLucqs: this pr is useless yes, i just wanted to see if we could get the signature with the right format

PR #10, opened 2026-02-05
Staking

micbakos:
## Summary

This PR adds staking functionality to the SDK and introduces a mobile example application.

### Staking API
- Implemented staking module with support for pool and staker operations
- Added validator presets for mainnet and Sepolia
- Added pool token handling and active pool tokens support
- Integrated with minting curve, reward supplier, and staking contracts

### Integration Testing
- Added integration tests for staking functionality using **forked Sepolia**
- Added global setup for integration test environment (ditched sh script to start devnet and leveraged https://github.com/0xSpaceShard/starknet-devnet-js

### Mobile Example
- Added React Native/Expo example app demonstrating SDK usage
- Includes screens for **balances**, **transfers**, and **staking**
- Features wallet connection, token management, and staking position tracking

### Other Changes
- Added new ABI files for staking contracts (pool, staking, minting-curve, reward-supplier)
   - minting-curve, reward-supplier are not used ATM, `getAPY()` on a staker/pool function is TBD  
- Added `generate-validator-presets.ts` script (fetching from voyager). In the future when [`get_stakers`](https://github.com/starkware-libs/starknet-staking/blob/main/docs/spec.md#get_stakers) function exists in contract we could allow the user to fetch that information and not rely on indexers.

In order to test, make sure to 

1. delete all `node_modules`,
2. `npm install` (on x)
3. `npm run build`
4. `cd examples/mobile`
5. `npm install`
6. `npx expo start --clear`


src/staking/staking.ts
```diff
@@ -0,0 +1,738 @@
...
+    amount: Amount,
+    options?: ExecuteOptions
+  ): Promise<Tx> {
+    await this.assertIsMember(wallet);
+
+    const tokenContract = this.tokenContract(wallet.getAccount());
+    const approveCall = tokenContract.populateTransaction.approve(
+      this.pool.address,
+      amount.toBase()
+    );
```
0xLucqs, line 221: `approve` could be added in the erc20 class no ?

micbakos: If I put this on erc20, then it will return a `Call`, or will it `execute`? Because the purpose of this is to create a multicall.

src/erc20/erc20.ts
```diff
@@ -0,0 +1,110 @@
...
+   * console.log(balance.toUnit());      // "100.5"
+   * console.log(balance.toFormatted()); // "100.5 USDC"
+   * ```
+   */
+  public async balanceOf(wallet: WalletInterface): Promise<Amount> {
+    const provider = wallet.getProvider();
+    const address = wallet.address;
+    const result = await provider.callContract({
+      contractAddress: this.token.address,
+      entrypoint: "balanceOf",
```
0xLucqs: call either `balanceOf` or `balance_of` Some contracts only have 1 of the 2 functions

src/staking/staking.ts
0xLucqs: good point. Long term i guess we want a "transaction builder" or something like that so we can aggregate the calls. Something like `wallet.approve().transfer().stake().unstake().balance_of().execute()` So let's ignore that for now and have a dedicated pr WDYT ?

micbakos: Yes I agree, the builder approach sounds very nice!

PR #12, opened 2026-02-06
fix: braavos deployment with paymaster

0xLucqs:


src/wallet/index.ts
```diff
@@ -29,6 +30,21 @@ import {
...
+
+// Braavos factory address (same on Sepolia and Mainnet)
+const BRAAVOS_FACTORY_ADDRESS =
+  "0x3d94f65ebc7552eb517ddb374250a9525b605f25f4e41ded6e7d7381ff1c2e8";
+
+// Chain IDs as felt252
+const CHAIN_ID_FELTS: Record<string, string> = {
+  SN_MAIN: "0x534e5f4d41494e",
+  SN_SEPOLIA: "0x534e5f5345504f4c4941",
+};
```
micbakos: Should we put this into `ChainId` definition?

0xLucqs: added a chainId type

PR #50, opened 2026-03-02
Decouple native

micbakos:
Attempt to separate packages so native clients install `@starkzap/native`. The reasoning behind this is to integrate differently when integrations separate when running on native (like cartridge).

* Core sdk exposes cartridge as optional peer dependency. The client needs to install it separately.
* Core sdk sets up cartridge for web by default
* Native sdk provides different implementation for setting up cartridge (either with connect or onboard).
* Provides specific polyfills and the installation is as easy as
```
// Install Starkzap Native polyfills first
import "@starkzap/native/polyfills";
```
* Updated examples.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,32 +52,18 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
-          cache-dependency-path: |
-            package-lock.json
-            examples/web/package-lock.json
-            examples/flappy-bird/package-lock.json
-            examples/mobile/package-lock.json
 
-      - name: Install Root Dependencies
+      - name: Install (workspaces)
         run: npm ci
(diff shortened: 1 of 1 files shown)


examples/mobile/stores/wallet.ts
```diff
@@ -246,18 +246,12 @@ export const useWalletStore = create<WalletState>((set, get) => ({
         contract: fromAddress(
           "0x00ca1702e64c81d9a07b86bd2c540188d92a2c73cf5cc0e508d949015e7e84a7"
         ),
-        mintingCurveContract: fromAddress(
-          "0x00ca1705e74233131dbcdee7f1b8d2926bf262168c7df339004b3f46015b6984"
```
ZackAttax, line 250: DRY!

PR #51, opened 2026-03-02
feat(payment): add Chainrails cross-chain payment module

GoodyBoy301:
## Summary

This PR introduces the Payment module powered by ChainRails, enabling cross-chain, multi-token payment acceptance for StarkZap applications.

### Features Added

- **Payment Session Management**: Create payment sessions for recipients with specific token and amount preferences
- **Quote Discovery**: Get quotes from all available bridges or find the best quote across bridges
- **Payment Intents**: Create, manage, and track payment intents with full lifecycle support
- **Cross-Chain Routing**: Find optimal routes and supported bridges for any source → destination chain pair
- **Chain & Balance Queries**: Query supported chains and token balances across chains
- **Payment Modal**: Web-based payment modal using ChainRails vanilla web component
- **Merchant Info**: Retrieve client/merchant account information

### Technical Details

- Integrated `@chainrails/sdk` (v0.3.3) and `@chainrails/vanilla` (v0.0.6)
- Added `payment` config option to `SDKConfig` with `apiKey` and optional `environment`
- Browser runtime detection for graceful fallback when payment is not configured
- Full TypeScript support with comprehensive type definitions

#### Get Quotes and Create Intent
```ts
// Get all quotes for a destination
const quotes = await payment.getAllQuotes({
  destinationChain: "STARKNET",
  tokenOut: "0x053c91...",
  amount: "10",
  recipient: "0xabc...",
});

// Create a payment intent
const intent = await payment.createIntent({
  sender: "0xsender...",
  amount: "10",
  tokenIn: "0xtokenIn...",
  amountSymbol: "USDC",
  source_chain: "BASE",
  destination_chain: "STARKNET",
  recipient: "0xrecipient...",
  refund_address: "0xsender...",
});
```

#### Payment Modal (Web)
```ts
// Create a modal handle with the session token
const modalHandle = payment.modal({
  sessionToken: session.sessionToken,
  amount: "25.00",
  platform: "web", // defaults to "web"
});

// Trigger the payment modal - returns true on success, false on cancel
const success = await modalHandle.pay();

if (success) {
  console.log("Payment completed successfully!");
} else {
  console.log("Payment was cancelled or failed");
}
```

eg. A variant of flappy-bird where the rich live forever
https://github.com/user-attachments/assets/83ea0649-e7aa-4790-9d2a-50a8401ea525

### Tests
Added comprehensive test suite with 24 test cases covering:
- Payment constants (chains, tokens, bridges)
- Session creation
- Modal handling
- Quote fetching (all, best, single bridge)
- Intent management (create, get, list, trigger processing)
- Router operations
- Chain queries
- Client info retrieval
- SDK integration

### Breaking Changes
None - Payment module is optional and only activated when payment config is provided.

### Related Dependencies
@chainrails/sdk: ^0.3.3
@chainrails/vanilla: ^0.0.6

## Summary by CodeRabbit

* **New Features**
  * Cross-chain, multi-token payments with session creation and a web payment modal; SDK adds a payment() surface.

* **Examples**
  * Added a Next.js demo and updated the Flappy Bird demo with an in-app "Save Me" revive checkout using the payment flow.

* **Documentation**
  * Full payment configuration and usage guides; README clarifies browser vs server payment behavior.

* **Tests**
  * Comprehensive test suite covering the payment surface and SDK integration.

* **Chores**
  * Dev tooling: pre-commit checks, lint/prettier ignores, build script tweaks, env/gitignore scaffolding.


examples/chainrails-nextjs-example/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### 2. Set Environment Variables
+
+Create a `.env.local` file in the root directory:
+
+```
+NEXT_PUBLIC_API_URL=http://localhost:3000
+NEXT_PUBLIC_CHAINRAILS_API_KEY=your_api_key_here
+```
```
0xLucqs: The documented setup tells users to set `NEXT_PUBLIC_CHAINRAILS_API_KEY`, but the API route actually reads `CHAINRAILS_API_KEY`. Following the README and `.env.example` as written leaves the server endpoint without the credential it needs to create sessions.

src/sdk.ts
```diff
@@ -478,6 +479,51 @@ export class StarkZap {
...
+   *
+   * // Create an intent
+   * const intent = await payment.createIntent({ ... });
+   * ```
+   */
+  payment(): Payment {
+    if (!this.config.payment) {
+      if (isWebRuntime()) {
+        return new Payment({ apiKey: "" });
+      }
```
0xLucqs, line 617: The browser fallback path returns `new Payment({ apiKey: "" })`, and the `Payment` constructor always re-runs `Chainrails.config(...)`. Because Chainrails keeps auth in a global singleton, hitting this path after a configured instance wipes the real API key for subsequent authenticated requests in the same runtime.

src/payment/payment.ts
```diff
@@ -0,0 +1,379 @@
...
+    return crapi.quotes.getAll(input);
+  }
+
+  /**
+   * Get quotes for the current session (requires prior `createSession` call).
+   */
+  async getSessionQuotes(
+    input: GetSessionQuotesInput
+  ): Promise<GetSessionQuotesOutput> {
+    return crapi.quotes.getAllForSession(input);
```
0xLucqs, line 264: `createSession()` returns a session token, but the class never stores it and the later session helpers (`getSessionQuotes()`, `createSessionIntent()`, `getSessionClientInfo()`, etc.) accept no session token parameter. That makes the documented `requires prior createSession call` flow impossible through this wrapper alone; a local repro confirmed `createSession()` leaves Chainrails stored session token empty.

examples/chainrails-nextjs-example/src/app/page.tsx
```diff
@@ -0,0 +1,325 @@
...
+import NGFlagIcon from "../icons/NGFlag";
+
+export default function Home() {
+  const [loading, setLoading] = useState(false);
+  const [paymentMethod, setPaymentMethod] = useState("credit-card");
+  // Initialize SDK and payment modal
+  const sdk = new StarkZap({
+    network: "mainnet",
+  });
+  const payment = sdk.payment();
```
0xLucqs: This component eagerly calls `sdk.payment()` during render. `next build` prerenders `/` in Node, so this code runs outside a browser and throws `Payment is not configured...`, which aborts the example build.

package-lock.json
```diff
@@ -54,6 +56,38 @@
...
+        "postcss": "^8.5.6",
+        "react": "^19.2.1",
+        "react-dom": "^19.2.1",
+        "starkzap": "file:../..",
+        "tailwindcss": "^4.1.18"
+      },
+      "devDependencies": {
+        "@types/node": "^20",
+        "@types/react": "^19",
+        "@types/react-dom": "^19",
```
0xLucqs: The new `examples/chainrails-nextjs-example` workspace declares `@types/react-dom`, but the root lockfile does not contain the matching resolved package entry. On this head, `npm ci --ignore-scripts` fails immediately with `Missing: @types/react-dom@19.2.3 from lock file`, so CI stops before lint, test, or build.

src/payment/types.ts
```diff
@@ -0,0 +1,266 @@
...
+export type GetChainBalanceInput = Parameters<
+  typeof crapi.chains.getBalance
+>[0];
+
+// ─── Auth / Session types ──────────────────────
+
+/** Input for creating a payment session. */
+export type CreatePaymentSessionInput = Parameters<
+  typeof crapi.auth.getSessionToken
+>[0];
```
0xLucqs, line 323: `CreatePaymentSessionInput` is re-exported directly from `crapi.auth.getSessionToken`, whose current type narrows `token` to the literal `"USDC"`. That means the SDK docs around session creation with other tokens do not type-check for consumers of this PR.

src/payment/payment.ts
```diff
@@ -0,0 +1,379 @@
...
+   */
+  modal(input: PaymentModalInput): PaymentModalHandle {
+    const platform = input.platform ?? "web";
+
+    const handle: PaymentModalHandle = {
+      platform,
+      sessionToken: input.sessionToken,
+      pay: async () => {
+        const modalManager = await this.getModalManager();
+        return modalManager.modal({ ...input, platform }).pay();
```
0xLucqs: `.pay()` imports `@/payment/modal` before proving that it is in a browser. That module immediately loads `@chainrails/vanilla`, so a Node caller gets a package import failure (`Unknown file extension ".css"`) instead of the intended `requires a browser environment` SDK error.

examples/chainrails-nextjs-example/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "starkzap-nextjs-demo",
```
0xLucqs: This PR introduces a brand-new example workspace (`examples/chainrails-nextjs-example`) instead of extending one of the existing examples. If the agreed scope is to keep example work within the current demos, this needs to be folded into an existing example or dropped from the PR.

GoodyBoy301: initially added the example, cos it showed a usage in react, figured most real-life usacases would be within a framework. Dropped it from the PR

src/payment/payment.ts
GoodyBoy301: moved the @chainrails/vanilla to a dynamic import, and the pay method confirms it's used on a browser before proceeding

src/payment/types.ts
GoodyBoy301: updated the docs to reflect only USDC settlements being supported currently

src/payment/payment.ts
GoodyBoy301: Created sessionTokens are now stored. Also added a sessionToken setter, for manual override

src/sdk.ts
GoodyBoy301: Modified the logic to only wipe the apikey if there's a new one

mintlify-docs/build/consumer-app-sdk/payment.mdx
0xLucqs, line 1: Several examples in this guide are out of sync with the actual wrapper shapes: `getAllQuotes()` returns an object with `quotes` and `cheapestOption`, not a bare array; `getBestQuote()` exposes `bridgeToUse`, not a top-level `bridge`; `getSupportedBridges()` returns `{ supportedBridges, routeInfo }`; and `getClientInfo()` uses `id`, not `clientId`. A consumer following these snippets will hit type errors immediately, so the docs should mirror the exported types before this ships.

src/payment/types.ts
```diff
@@ -0,0 +1,266 @@
...
+export type GetSessionQuotesOutput = AsyncResult<
+  typeof crapi.quotes.getAllForSession
+>;
+
+// ─── Intent types ──────────────────────────────
+
+/** A fully resolved payment intent. */
+export type PaymentIntent = Intent;
+
+/** Input for creating a payment intent. */
```
0xLucqs: `GetAllQuotesInput` and the other quote helpers expose SDK-style camelCase fields, but `CreatePaymentIntentInput` and `PaymentIntent` are re-exported directly from Chainrails and force consumers back to snake_case names like `source_chain`, `destination_chain`, `refund_address`, and `intent_status`. That makes the payment module much harder to learn than the rest of StarkZap because users have to switch naming conventions mid-flow. If this surface is meant to feel SDK-native, the public intent request/response types should be normalized here instead of leaking the backend contract.

0xLucqs:
@Darlington02 @GoodyBoy301 what's the state of this ?

GoodyBoy301:
> @Darlington02 @GoodyBoy301 what's the state of this ?

Rounding up the mobile support, then it would it be ready for review again.
Would drop a comment as soon as it is, later today

GoodyBoy301:
Hi @0xLucqs , this is ready to be reviewed

0xLucqs (changes requested):
## Senior-engineer review — request changes

Reviewed the payment module in depth (~1,700 LoC of actual changes, ignoring the merge noise). There's a solid foundation, but several blockers prevent merge as-is. Summary: `Session.create` is broken end-to-end, the test file contains 5 TypeScript errors and 3 real assertion failures, one source file is a dead duplicate, and a silent-init path leaves the `Payment` instance usable but non-functional. Running CI locally reproduces the failures — and the PR's own `test` and `examples` jobs are red.

### Verification methodology

1. `diff src/payment/session.ts src/payment/intents.ts` — one-line difference.
2. `tsc --noEmit` with `include: ["src/**/*", "tests/**/*"]` — 5 errors in `tests/payment.test.ts`.
3. `vitest run tests/payment.test.ts` — 3 tests fail (`createSession`, `createIntent`, `getIntent`).
4. Traced the `session.create → server.ts → flappy-bird main.ts` flow against Chainrails's own `.d.ts` types.
5. Cross-checked CI: `gh pr checks 51` shows `test` and `examples` jobs failing.

---

## 🔴 Critical (blocks merge)

### C1 — `tests/payment.test.ts` doesn't typecheck and 3 tests actually fail

The project's `npm run typecheck` only covers `src/**/*` (see `tsconfig.json#include`), so the test file's type errors are hidden. Adding `tests/**` to the tsc include surfaces:

```
tests/payment.test.ts(24,3):  TS2322 Type 'number' is not assignable to type 'string'.
tests/payment.test.ts(140,21): TS2551 Property 'sessionToken' does not exist on type 'Payment'. Did you mean 'setSessionToken'?
tests/payment.test.ts(141,21): TS2339 Property 'amount' does not exist on type 'Payment'.
tests/payment.test.ts(258,9):  TS2561 'source_chain' does not exist in type 'CreatePaymentIntentInput'. Did you mean 'sourceChain'?
tests/payment.test.ts(284,21): TS2551 Property 'intent_status' does not exist on type 'PaymentIntent'. Did you mean 'intentStatus'?
```

`MOCK_INTENT` (`tests/payment.test.ts:23-60`) is annotated `PaymentIntent` but uses the Chainrails snake_case shape (`id: 1`, `client_id`, `source_chain`, `intent_status`, etc.). `createIntent` test input passes snake_case keys that don't exist on `CreatePaymentIntentInput`. Assertions check `result.intent_status` on a camelCase-normalised return.

Running the file confirms it's not just TS: **3 tests fail at runtime** (`createSession`, `createIntent`, `getIntent`). The `as never` casts on every `mockResolvedValue(... as never)` silence the type system but not the runtime assertions.

**Fix:** rewrite `MOCK_INTENT` and `createIntent` input to match the camelCase interfaces. Then **add `tests/**` to the `tsc --noEmit` run** (or make a tests-only `tsconfig.test.json`) so this can't silently regress. Vitest is transpile-only — without tsc over tests, drift goes unnoticed.

### C2 — `Session.create()` return type is wrong; breaks the server + flappy-bird example

`src/payment/session.ts:11-15` returns `Promise<Payment>` after stashing the token internally. Chainrails's own type says `Auth.getSessionToken` returns `{ sessionToken, amount }` (see `@chainrails/sdk/dist/src/Auth/index.d.ts`).

Downstream consequences:

- `tests/payment.test.ts:140-141` expects `result.sessionToken === "tok_123"` and `result.amount === "25.00"` — both are `undefined` because `result` is a `Payment` instance. Test fails.
- `examples/server/server.ts:215-223` does `res.json(session)` — serialises the `Payment` instance, which has `currentSessionToken` (not `sessionToken`) and an embedded `session` sub-object. The HTTP body is `{ currentSessionToken: "...", session: {} }`.
- `examples/flappy-bird/main.ts:43-47` reads `payload.sessionToken` on that response → `undefined` → throws `"Session payload is missing a valid sessionToken."` **The Save-Me demo cannot work.**

**Fix:** make `Session.create` return `Promise<{ sessionToken: string; amount: string }>` (the Chainrails response) and keep the `setSessionToken` side-effect. Update test, server example.

### C3 — `src/payment/intents.ts` is a dead duplicate of `session.ts`

`diff src/payment/session.ts src/payment/intents.ts` shows one line of difference (`private parent: Payment;` vs `parent: Payment;`). Grep confirms **zero** imports of `@/payment/intents` anywhere in `src/`. Delete the file before a contributor edits the wrong one and introduces drift.

### C4 — `sdk.payment()` silent-init lets misconfigured `Payment` escape

`src/sdk.ts:609-619` returns `new Payment({ apiKey: "" })` when `payment` config is missing in browser/native runtimes. `Payment` constructor (`src/payment/payment.ts:114`) then **skips** `Chainrails.config(...)` because `config.apiKey` is falsy.

I suspect the intent is to support the "server-minted sessionToken + client-side `checkout()`" flow (which legitimately doesn't need an apiKey on the client). That's reasonable — but every *other* method on the returned `Payment` (`createIntent`, `getQuotes`, ramp, …) will fail deep inside `@chainrails/sdk` with a confusing auth error, not a clear SDK-level one.

**Fix:** add a `configured: boolean` flag and have every non-`checkout` method throw a clear "Payment was not configured with an apiKey — either pass `payment: { apiKey }` to `new StarkZap()` or only use `checkout()` with a server-minted session token" error.

### C5 — `void Chainrails.config(...)` is fire-and-forget on a global singleton

`src/payment/payment.ts:115-118` drops the promise. Verified against the vendor type: `Chainrails.config()` returns `Promise<Chainrails>` and the class uses `private static app` — one shared process-global. Consequences:

- Config errors (network, invalid key) become unhandled rejections.
- Every subsequent `crapi.*` call races the config promise; timing-dependent failures.
- Constructing two `StarkZap` instances with different keys: last one to instantiate steals auth for everyone.

`sdk.payment()` (see M8 below) makes this worse by constructing a fresh `Payment` on every call — the example server re-runs config on every `/session-token` request under concurrent load.

**Fix:** store `configPromise` on `Payment` and `await` it at the top of every method that hits `crapi.*`. Cache `Payment` as a lazy singleton on `StarkZap`.

### C6 — `checkout.pay()` has no reject path — the caller hangs forever on failure

`src/payment/checkout.ts:26` — `new Promise<boolean>((resolve) => ...)` accepts no reject. The promise resolves only via `onSuccess`/`onCancel` callbacks. If any of the following happens, `await handle.pay()` never settles:

- `await import("@chainrails/vanilla")` rejects (CSP, network, bundle missing).
- `document.createElement(...)` returns an element without `setProps` — `modal.setProps?.(...)` silently no-ops at line 65, then `modal.open()` shows an empty modal that never calls `onSuccess`/`onCancel`.
- `modal.open()` throws (not wrapped).

Zero telemetry, zero user feedback.

**Fix:** accept `reject`, wrap `modal.open()` in try/catch, guard `setProps` with an explicit error on absence, and consider a timeout.

### C7 — CI is actually failing (the summary is misleading)

`gh pr checks 51` reports:

- `test` — **fail** (2m31s)
- `examples` — **fail** (4m42s)
- `CodeRabbit` — pass

The `test` failure is C1/C2 reproduced in CI; the `examples` failure is most likely M6 below. Please confirm green CI before requesting review.

---

## 🟠 Important

### I1 — Unsafe union cast on `amountSymbol`
`src/payment/payment.ts:298-308` casts `input.amountSymbol as "USDC" | "USDT" | ...`, but `CreatePaymentIntentInput.amountSymbol` is typed `string` (`src/payment/types.ts:245`). Any caller passing `"FOO"` dies downstream. Type it as `PaymentTokenSymbol` (already exported, backed by `AmountSymbols`) and drop the cast. Per CLAUDE.md: *"prefer explicit domain types over primitives."*

### I2 — Unsafe `as Parameters<typeof ...>[0]` cast
`src/payment/payment.ts:250-255` — rewrites input, then force-asserts the shape. Defeats strict typing. Reflect Chainrails's actual requirements in `GetQuotesFromAllBridgesInput` rather than rewriting at the call site.

### I3 — `createIntent` duplicates `normalizeIntent`
`src/payment/payment.ts:292-336` inlines the snake→camel conversion already available in `normalizeIntent` (`:151-183`). Worse, `normalizeIntent` casts `intent_status` to the union but `createIntent` writes `result.intent_status` unfiltered. Collapse to `return this.normalizeIntent(await crapi.intents.create(snakeCaseInput))`. `createSessionIntent` already does this correctly.

### I4 — `.map(this.normalizeIntent)` loses `this` binding
`src/payment/payment.ts:369, 381, 391`. Works today because `normalizeIntent` doesn't read `this`. Brittle: silently breaks if someone adds `this.*` inside the method. Use `.map((r) => this.normalizeIntent(r))` or declare `normalizeIntent` as a static/arrow-bound method.

### I5 — `listIntents` mutates upstream response
`src/payment/payment.ts:378-383`: `result.intents = result.intents.map(...)`. If `crapi` caches internally (plausible), this pollutes its cache. Build a new object: `return { ...result, intents: result.intents?.map(this.normalizeIntent) }`.

### I6 — Orphan JSDoc for a nonexistent method
`src/payment/payment.ts:189-207` documents `createSession` (which lives on `Session`, not `Payment`) and sits above `checkout()` which has its own doc at `:209-215`. Looks like a merge leftover. Delete or move to `Session.create`.

### I7 — `payment-modules.d.ts` under-declares the modal element
`src/types/payment-modules.d.ts` declares `ChainrailsPaymentModalElement: { tagName: string }`, but `src/payment/checkout.ts:27-40` casts to `HTMLElement & { setProps?, open, close }` inline. The stub hides the real shape and enables the optional-chain silent no-op from C6. Either declare the real API in the `.d.ts` or remove the stub and use Chainrails's own types.

### I8 — Empty catch in flappy-bird example
`examples/flappy-bird/main.ts:99-101` — `try { ... } catch { closeSaveMeModal(false); }`. Collapses network failures, malformed JSON, SDK exceptions, and user cancellation into one outcome with no logging. Examples are copy-paste templates: `catch (err) { console.error("save-me payment failed", err); closeSaveMeModal(false); }`.

### I9 — `/session-token` endpoint logs nothing server-side
`examples/server/server.ts:224-227` — sends 500 to the client but the adjacent `/api/paymaster` handler at `:199-203` does log. Add `console.error("[SessionToken] Exception:", message)` for consistency.

### I10 — `assertBrowser()` fires late
`src/payment/checkout.ts:17` — runs inside `.pay()`, so server-side construction of a handle succeeds and only blows up at pay time. Minor; worth moving to `sdk.payment().checkout(...)` for earlier failure.

---

## 🟡 Minor

- **M1 — dead `PaymentModalPlatform` / `.platform`** (`src/payment/types.ts:95-103`). Declared, accepted, never read. Wire up a mobile branch or delete.
- **M2 — several fields use `T | undefined` instead of `T?`** (`src/payment/types.ts:221-230`). With `exactOptionalPropertyTypes`, callers must pass `undefined` explicitly. Use `metadata?: {...}`, `createdAt?: string`, `expiresAt?: string`.
- **M3 — inconsistent amount defaulting**. `src/payment/payment.ts:225-227` omits `handle.amount` when absent; `src/payment/checkout.ts:10, 67` always forces `input.amount || "0"`. The `"0"` fabrication is a real trap if a UI ever passes `""`. Pick one — I'd prefer "absent when absent" everywhere.
- **M4 — leaky generic ramp return types**. `src/payment/payment.ts:598-638` — `AsyncResult<typeof crapi.ramp.xxx>` bleeds the vendor's exact types across the SDK boundary. A minor field rename in Chainrails silently breaks consumers. Declare stable shapes in `types.ts`.
- **M5 — inconsistent `Chainrails` vs `crapi` import**. Constructor uses `Chainrails.config()`, everything else uses `crapi.*`. Pick one or document why.
- **M6 — `examples/tic-tac-toe/package.json` adds a 17-entry `peerDependencies` block** that didn't exist before the PR. Examples are leaf consumers; peer deps on `@radix-ui/react-compose-refs`, `@react-navigation/core`, `@expo/metro-runtime`, etc. should not be declared here. Almost certainly the root cause of the `examples` CI failure. Revert.
- **M7 — `packages/native/src/index.ts` switched from `@/sdk` → `starkzap`** and `tsconfig.json` now has `"noEmit": false`. This requires a pre-built `dist/` for the native package to typecheck. Document in `packages/native/README.md`.
- **M8 — `sdk.payment()` returns a fresh `Payment` every call** (`src/sdk.ts:618`). Should be a memoised lazy singleton — see C5 for the race hazard.
- **M9 — inconsistent server startup gating** (`examples/server/server.ts:32-41`). `PRIVY_APP_ID`/`AVNU_API_KEY` missing → `process.exit(1)`. `CHAINRAILS_API_KEY` missing → `console.warn` + 503 on call. Pick one policy.

---

## ✅ What's good

- `src/payment/types.ts` is well-structured, with clear sectioning, JSDoc on most types, and a sensible split between re-exports from `@chainrails/sdk` (`AmountSymbols`, `Bridges`, `Chains` as values) and SDK-native types.
- Lazy dynamic `import("@/payment/checkout")` via `getModalManager()` correctly keeps `@chainrails/vanilla` out of server bundles.
- `PaymentModalManager` tracks `activeModalCleanup` and uses a `settled` flag to make `pay()` idempotent — correct shape, just needs the reject path (C6).
- Broad Chainrails surface covered: quotes, intents, router, chains, client info, ramp.
- `assertSafeHttpUrl` is properly reused for RPC/explorer URL validation elsewhere in the SDK.

---

## Recommended fix order

1. Delete `src/payment/intents.ts` (C3 — trivial).
2. Fix `Session.create` return type + test + server example (C2).
3. Rewrite `MOCK_INTENT` and `createIntent` test inputs to match the camelCase interfaces, then **add `tests/**` to the typecheck** (C1). Confirm `vitest run` is green.
4. Add reject path to `checkout.pay()` and explicit-error guard for `setProps` (C6).
5. Harden `sdk.payment()`: either throw on non-`checkout` methods when apiKey empty, or cache a singleton and memoise a `configPromise` (C4 + C5).
6. Tighten `amountSymbol` to `PaymentTokenSymbol`, remove `as Parameters<>[0]` cast, collapse `createIntent` into `normalizeIntent` (I1 + I2 + I3).
7. Fix remaining I-series.
8. Revert `examples/tic-tac-toe/package.json` peerDependencies block (M6) — likely unblocks the `examples` CI job.

GoodyBoy301:
Hi @0xLucqs , made a couple more updates, based on the review

Darlington02:
Hi @0xLucqs popping this up

0xLucqs:
Hello, Thanks for the work but there are still remaining bugs here and i think even the upstream chainrails sdk has bugs. We won't have time to fix everything before the release (we're aiming Thursday).
We'll re assess chainrails for v4 later next week

PR #62, opened 2026-03-07
add tongo to starkzap

adrienlacombe:
## Summary by CodeRabbit

* **New Features**
  * Confidential account operations added: fund, transfer, withdraw, ragequit, rollover; available in the SDK and usable via the transaction builder for batched workflows.
* **Types**
  * Public types/interfaces added for confidential account configuration, operation details, and account state.
* **Chores**
  * Added runtime dependency for Tongo SDK integration.
* **Tests**
  * Extensive unit and integration tests (including local devnet scenarios) covering confidential flows.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,13 +14,13 @@
       ],
       "dependencies": {
         "@avnu/avnu-sdk": "^4.1.0-rc.0",
-        "@fatsolutions/tongo-sdk": "^1.3.2",
         "starknet": "^9.2.1"
       },
       "devDependencies": {
         "@0no-co/graphql.web": "^1.2.0",
         "@cartridge/controller": "^0.13.9",
         "@eslint/js": "^9.39.2",
+        "@fatsolutions/tongo-sdk": "^1.3.2",
         "@vitest/coverage-v8": "^4.0.17",
         "bestzip": "^2.2.1",
         "dotenv": "^17.2.3",
@@ -40,13 +40,17 @@
         "vitest": "^4.0.17"
       },
       "peerDependencies": {
-        "@cartridge/controller": "^0.13.9"
+        "@cartridge/controller": "^0.13.9",
+        "@fatsolutions/tongo-sdk": "^1.3.2"
       },
       "peerDependenciesMeta": {
         "@cartridge/controller": {
           "optional": true
         },
         "@ethersproject/shims": {
+          "optional": true
+        },
+        "@fatsolutions/tongo-sdk": {
           "optional": true
         },
         "fast-text-encoding": {
@@ -458,6 +462,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -1884,6 +1889,7 @@
       "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.28.6.tgz",
       "integrity": "sha512-05WQkdpL9COIMz4LjTxGpPNCdlpyimKppYNoJ5Di5EUObifl8t4tuLuUBBZEpoLYOmfvIWrsp9fCl0HoPRVTdA==",
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=6.9.0"
       }
@@ -2979,7 +2985,8 @@
           "url": "https://www.buymeacoffee.com/ricmoo"
         }
       ],
-      "license": "MIT"
+      "license": "MIT",
+      "peer": true
     },
     "node_modules/@expo/cli": {
       "version": "54.0.23",
@@ -4080,6 +4087,7 @@
       "version": "0.4.0",
       "resolved": "https://registry.npmjs.org/@fatsolutions/she/-/she-0.4.0.tgz",
       "integrity": "sha512-zPwERSMQ1w0QmoSyhSqCApAeyG8WLRkefsk/alHzx9OeZ503d8pGbNg7DDJBQ/UyfqCKevs6sjql6/k/cE/jaw==",
+      "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@noble/ciphers": "1.3.0",
@@ -4093,6 +4101,7 @@
       "version": "1.3.0",
       "resolved": "https://registry.npmjs.org/@noble/ciphers/-/ciphers-1.3.0.tgz",
       "integrity": "sha512-2I0gnIVPtfnMw9ee9h1dJG7tp81+8Ob3OJb3Mv37rx5L40/b0i7djjCVvGOVqc9AEIQyvyu1i6ypKdFw8R8gQw==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": "^14.21.3 || >=16"
@@ -4105,6 +4114,7 @@
       "version": "1.7.0",
       "resolved": "https://registry.npmjs.org/@noble/curves/-/curves-1.7.0.tgz",
       "integrity": "sha512-UTMhXK9SeDhFJVrHeUJ5uZlI6ajXg10O6Ddocf9S6GjbSBVZsJo88HzKwXznNfGpMTRDyJkqMjNDPYgf0qFWnw==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@noble/hashes": "1.6.0"
@@ -4120,6 +4130,7 @@
       "version": "1.6.0",
       "resolved": "https://registry.npmjs.org/@noble/hashes/-/hashes-1.6.0.tgz",
       "integrity": "sha512-YUULf0Uk4/mAA89w+k3+yUYh6NrEvxZa5T6SY3wlMvE2chHkxFUUIDI8/XW1QSC357iA5pSnqt7XEhvFOqmDyQ==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": "^14.21.3 || >=16"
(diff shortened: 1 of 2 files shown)


tests/integration/confidential.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+const TONGO_STATE_DIR = path.resolve(__dirname, "tongo-state");
+const tongoMeta = JSON.parse(
+  readFileSync(path.join(TONGO_STATE_DIR, "devnet.json"), "utf-8")
+) as {
+  contracts: {
+    Tongo: { address: string };
+    STRK: { address: string };
+  };
+};
```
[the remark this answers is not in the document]
adrienlacombe: won't fix

tests/integration/tongo-state/regenerate.mjs
```diff
@@ -0,0 +1,161 @@
...
+);
+let casm;
+try {
+  casm = json.parse(readFileSync(casmPath, "utf-8"));
+} catch {
+  console.log("No CASM file found, devnet will compile Sierra to CASM");
+}
+
+const STRK_ADDRESS =
+  "0x4718F5A0FC34CC1AF16A1CDEE98FFB20C31F5CD61D6AB07201858F4287C938D";
```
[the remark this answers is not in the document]
adrienlacombe: fixed by https://github.com/keep-starknet-strange/starkzap/pull/62/commits/93aef8261e38a294aae2573bf3ade5649faa69fb

adrienlacombe:
@0xLucqs updated with Amount type, thank you for this!

src/confidential/tongo.ts
```diff
@@ -0,0 +1,186 @@
...
+    details: ConfidentialWithdrawDetails
+  ): Promise<Call[]> {
+    const op = await this.account.withdraw({
+      amount: details.amount.toBase(),
+      to: details.to,
+      sender: details.sender,
+      ...(details.feeTo !== undefined && { fee_to_sender: details.feeTo }),
+    });
+    return [op.toCalldata()];
+  }
```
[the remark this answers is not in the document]
adrienlacombe: won't fix:
1. No source of truth to validate against. ConfidentialConfig stores only privateKey, contractAddress, and provider — no token metadata. The Confidential class has no knowledge of the
  expected asset's decimals or symbol. To add this validation you'd need a breaking API change to the config, or an async on-chain call to fetch token info on every populate* call.          
  2. feeTo is already bigint, not Amount. The suggestion to validate feeTo as an Amount doesn't apply — it's a raw fee value that goes straight to the tongo-sdk.                             
  3. Amount is the validation layer. The whole point of switching from bigint to Amount is that the caller constructs the Amount with the correct token/decimals context. .toBase()           
  deterministically returns the raw bigint. If the caller passes an Amount with wrong decimals, that's a caller error — the same class of bug as passing the wrong bigint before.             
  4. No other SDK module does this. Looking at the existing patterns (ERC20, Staking), none validate Amount metadata inside populate* methods. The convention is that Amount carries its own
  context and .toBase() is trusted.            

  This finding proposes defensive validation with no available source of truth and would require either a breaking API change or expensive on-chain calls. Not fixing.

src/tx/builder.ts
```diff
@@ -415,6 +423,121 @@ export class TxBuilder {
...
+   * @param details - Rollover parameters (sender)
+   * @returns this (for chaining)
+   */
+  confidentialRollover(
+    confidential: Confidential,
+    details: ConfidentialRolloverDetails
+  ): this {
+    this.queueAsyncCalls(confidential.populateRollover(details));
+    return this;
+  }
```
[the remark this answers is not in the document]
adrienlacombe: won't fix:
No fix needed. Here's why:                                                                                                                                                                  
                                                                                                                                                                                              
  The sender in confidential operations is not the same concept as the wallet address in staking. In the tongo protocol, sender identifies the Starknet account that will submit the          
  transaction on-chain — which is often a relayer, not the wallet owner. This is an explicit design choice visible in the types:                                                              
                                                                                                                                                                                              
  /** Optional fee paid to sender (for relayed txs). */                                                                                                                                       
  feeTo?: bigint;                                                                         
                                               
  The feeTo field exists specifically for relayed transactions where sender differs from the wallet. Forcing sender = this.wallet.address would break the relayer pattern entirely.

  By contrast, staking methods use this.wallet.address because the staking protocol requires the pool member's own address — there's no relayer concept.

  The finding conflates two different sender semantics. Not fixing.

tests/integration/confidential.test.ts
```diff
@@ -0,0 +1,318 @@
...
+      );
+    }
+
+    return () => {
+      if (devnet) {
+        devnet.kill();
+        devnet = null;
+      }
+    };
+  }, 120_000);
```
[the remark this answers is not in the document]
adrienlacombe: No fix needed. Here's why:                                                                                                                                                                  
                                                                                                                                                                                              
  The confidential integration test cannot use the shared devnet. They serve fundamentally different purposes:                                                                                
                                                                                                                                                                                              
  - Shared devnet (globalSetup): starts fresh with --seed 0, no state dump, no deployed contracts. Tests use devnet_mint to fund accounts dynamically.                                        
  - Confidential devnet: loads a pre-built state dump (devnet.state) containing an already-deployed Tongo contract with specific class hash and address. This is required because the tongo   
  contract artifacts aren't available at test time — they come from a separate repo and are baked into the state dump via regenerate.mjs.                                                     
                                                                                                                                                                                              
  These can't be merged. The shared devnet has no Tongo contract, and adding one would require the tongo contract artifacts in CI, coupling two separate projects.                            
                                                                                                                                                                                              
  As for "fail fast instead of warn+skip": the graceful skip is intentional. If starknet-devnet isn't installed or the state dump is missing, unit tests and other integration tests should   
  still pass. This is standard for optional integration suites that depend on external tooling. The other integration tests (deploy, erc20, staking) also guard with isAlive() checks via the 
  shared setup — same pattern, different mechanism.

src/confidential/confidential.ts
```diff
@@ -0,0 +1,186 @@
...
+ *   provider: wallet.getProvider(),
+ * });
+ *
+ * // Fund confidential account
+ * const tx = await wallet.tx()
+ *   .add(...await confidential.populateFund({
+ *     amount: Amount.fromRaw(100n, token),
+ *     sender: wallet.address,
+ *   }))
+ *   .send();
```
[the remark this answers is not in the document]
adrienlacombe: addressed by https://github.com/keep-starknet-strange/starkzap/pull/62/commits/54251d11e322c8bc63aac7753af4398194e2e484

src/confidential/types.ts
```diff
@@ -0,0 +1,73 @@
...
+  /** Optional fee paid to sender (for relayed txs). */
+  feeTo?: bigint;
+}
+
+/** Details for a confidential transfer. */
+export interface ConfidentialTransferDetails {
+  /** Amount to transfer. */
+  amount: Amount;
+  /** Recipient's Tongo public key (as {x, y} coordinates). */
+  to: { x: BigNumberish; y: BigNumberish };
```
0xLucqs: This hard-codes the generic confidential abstraction back to Tongo. `ConfidentialProvider.address` is only a string, but transfers require a `{x,y}` public key object here, so code that only knows the `ConfidentialProvider` surface cannot transfer to another provider instance using the value the interface exposes. The integration test has to reach into `getTongoAccount().publicKey` for that reason. Can we move the recipient shape behind the provider abstraction before shipping this as a provider-agnostic API?

src/confidential/tongo.ts
```diff
@@ -0,0 +1,183 @@
...
+    return await this.account.tongoToErc20(confidentialAmount);
+  }
+
+  /**
+   * Build the Call for funding this confidential account.
+   *
+   * The caller is responsible for including an ERC20 approve call
+   * before this in the transaction batch.
+   */
+  async fund(details: ConfidentialFundDetails): Promise<Call[]> {
```
0xLucqs, line 112: `account.fund()` produces both the prerequisite approve call and the Tongo calldata, but this wrapper only returns `op.toCalldata()`. Because `TongoConfidential` never exposes which ERC20/token must be approved, consumers cannot actually build the documented `wallet.tx().approve(...).confidentialFund(...)` flow from the public API alone. The integration test has to bypass the wrapper via `getTongoAccount().fund()` to recover `op.approve`. Can we either expose the approval metadata on this type or include the prerequisite call here?

src/confidential/types.ts
adrienlacombe: https://github.com/keep-starknet-strange/starkzap/pull/62/commits/fad7fcc10da0db92de84f0cd89fce5f1b5552cbe

src/confidential/tongo.ts
adrienlacombe: going with inclusion

adrienlacombe: https://github.com/keep-starknet-strange/starkzap/pull/62/commits/8d7001eda12adfc653649461454ce6b36236abc3

examples/mobile/app/(tabs)/confidential.tsx
```diff
@@ -0,0 +1,932 @@
...
+      setIsRefreshing(false);
+    }
+  }, [confidential, addLog]);
+
+  const handleFund = useCallback(async () => {
+    if (!wallet || !confidential || !fundAmount.trim()) return;
+    setIsFunding(true);
+    addLog(`Funding confidential account with ${fundAmount}...`);
+    try {
+      const amount = Amount.parse(fundAmount.trim(), 18);
```
0xLucqs: This screen exposes multiple Tongo assets (`USDC`, `USDT`, `ETH`, `WBTC`, etc.), but the example still parses and formats every confidential amount with `18` decimals. That works for `STRK`/`ETH`, but it produces incorrect submitted amounts and wrong balance displays for non-18-decimal tokens like `USDC` and `USDT`. The web example already resolves the selected token metadata and uses that here; the mobile example should do the same for the balance formatting paths and the fund/transfer/withdraw flows.

adrienlacombe: sorry for that, I didn't know how to test the mobile app

adrienlacombe: https://github.com/keep-starknet-strange/starkzap/pull/62/commits/98166b373e62990e12d9712fba94f278fd31c294
Claude hardcoded the decimals, is that okay for a demo app?

PR #74, opened 2026-03-10
Bridge

micbakos:
# Cross-Chain Bridging Support

## Summary

Adds a full cross-chain bridging module to the Starkzap SDK, enabling developers to bridge tokens from **Ethereum** and **Solana** into Starknet. The feature supports multiple bridging protocols, handles external wallet connections, and exposes a unified API through the existing wallet interface -- so bridging feels like a natural extension of the SDK rather than a separate tool.

## What this enables

Developers using Starkzap can now:

- **Discover bridgeable tokens** via `sdk.getBridgingTokens(chain?)`, which queries the StarkGate tokens API and returns typed token objects for Ethereum and Solana
- **Connect external wallets** (MetaMask, Phantom, etc.) using the new `ConnectedEthereumWallet` and `ConnectedSolanaWallet` helpers, with built-in chain validation against the configured Starknet network
- **Deposit tokens to Starknet** through `wallet.deposit(recipient, amount, token, externalWallet)` -- the SDK automatically selects the correct bridge protocol based on the token
- **Query balances, allowances, and fee estimates** on the source chain before initiating a bridge transaction

## Supported bridge protocols

| Protocol | Source Chain | Use Case |
|----------|-------------|----------|
| **Canonical (StarkGate)** | Ethereum | Standard L1 to L2 deposits for ETH and most ERC20 tokens |
| **CCTP (Circle)** | Ethereum | Native USDC bridging with optional fast transfer |
| **OFT (LayerZero)** | Ethereum | LayerZero Omnichain Fungible Tokens (mainnet only, requires API key) |
| **Lords** | Ethereum | LORDS token with its own bridge contract |
| **Hyperlane** | Solana | Warp route bridging for Solana tokens to Starknet |

## Architecture

The bridge module follows a strategy pattern:

- **`BridgeInterface`** defines a protocol-agnostic contract (deposit, fee estimate, balance, allowance)
- **`BridgeOperator`** is the orchestrator that routes to the correct bridge implementation based on token metadata, caches bridge instances, and validates chain alignment
- **`EthereumBridge`** is an abstract base class providing shared logic (approval flows, gas estimation, error mapping) for all Ethereum bridge subclasses
- **`SolanaHyperlaneBridge`** is lazy-loaded to avoid bundling Node-only Solana dependencies when not needed

The bridge operator is embedded directly into the wallet, so `WalletInterface` now extends `BridgeOperatorInterface`. This means every wallet implementation (standard, Cartridge) automatically gains bridge capabilities.

## Configuration

A new optional `bridging` field on `SDKConfig` supports:

- `layerZeroApiKey` -- required only for OFT tokens
- `ethereumRpcUrl` -- custom RPC for gas estimation (falls back to the connected wallet's provider)
- `solanaRpcUrl` -- custom RPC for Solana (falls back to public cluster)

## External wallet connection

The new `src/connect/` module provides typed wrappers for external chain wallets:

- `ConnectedEthereumWallet.from(provider, address, chainId)` -- validates EIP-1193 provider and Ethereum address checksums, ensures the Ethereum network matches the configured Starknet chain (mainnet/sepolia)
- `ConnectedSolanaWallet.from(provider, address, chainId)` -- validates Solana public key and network alignment

Both integrate with **Reown AppKit** for multi-chain wallet connection in the examples.

## Token repository

`BridgeTokenRepository` fetches available bridge tokens from the StarkGate API with:

- In-memory caching (1-hour TTL)
- Request deduplication for concurrent queries
- Automatic filtering of hidden/deprecated tokens
- Environment inference from the configured Starknet chain ID

## Examples

- **Web example** (`examples/web/`) -- Full bridge UI with direction toggle, token picker, balance/allowance display, CCTP fast transfer checkbox, fee breakdown, and deposit flow
- **Mobile example** (`examples/mobile/`) -- React Native bridge tab with Ethereum/Solana chain selector, percentage amount buttons, fee display, and AppKit wallet integration (Phantom, Solflare, MetaMask)

## Tests

- Unit tests for `BridgeTokenRepository` covering token mapping, query parameter construction, caching behavior, and filtering of hidden/deprecated tokens

## Notable design decisions

- Bridge methods live on the wallet interface rather than as a standalone module, keeping the API surface consistent with existing SDK patterns (transfers, staking)
- The Solana Hyperlane bridge is lazy-loaded via dynamic import to avoid pulling in `@solana/web3.js` when only Ethereum bridging is needed
- Fee estimation returns structured breakdowns (L1 gas, L2 message fee, approval fee, CCTP percentage) so UIs can display granular cost information
- Bridge instances are cached per token+wallet combination and invalidated when the wallet changes

### DIFF
--- a/src/bridge/tokens/repository.ts
+++ b/src/bridge/tokens/repository.ts
@@ -127,6 +127,31 @@ const isNonNull = <T>(value: T | null): value is T => value !== null;
 type NormalizeEthereumAddress = (value: string) => EthereumAddress;
 type NormalizeSolanaAddress = (value: string) => SolanaAddress;
 
+function getTokenChain(token: BridgeTokenApiRecord): ExternalChain | null {
+  if (typeof token.chain !== "string") {
+    return null;
+  }
+
+  switch (token.chain.trim().toLowerCase()) {
+    case ExternalChain.ETHEREUM:
+      return ExternalChain.ETHEREUM;
+    case ExternalChain.SOLANA:
+      return ExternalChain.SOLANA;
+    default:
+      return null;
+  }
+}
+
+function isOptionalPeerDependencyError(
+  error: unknown,
+  dependency: "ethers" | "@solana/web3.js"
+): error is Error {
+  return (
+    error instanceof Error &&
+    error.message.includes(`optional peer dependency "${dependency}"`)
+  );
+}
+
 function parseToken(
   token: BridgeTokenApiRecord,
   normalizeEthereumAddress?: NormalizeEthereumAddress,
@@ -289,6 +314,7 @@ export class BridgeTokenRepository {
     query: BridgeTokenQuery,
     key: string
   ): Promise<BridgeToken[]> {
+    const isExplicitChainRequest = query.chain !== undefined;
     const url = new URL(this.apiUrl);
     url.searchParams.set("env", query.env ?? DEFAULT_ENV);
     if (query.chain) {
@@ -307,24 +333,61 @@ export class BridgeTokenRepository {
     }
 
     const payload = assertArrayPayload(await response.json());
-    const hasEthereumRows = payload.some((token) => {
-      return (
-        typeof token.chain === "string" &&
-        token.chain.toLowerCase() === ExternalChain.ETHEREUM
-      );
+    const visiblePayload = payload.filter((token) => {
+      return !token.hidden && !token.deprecated;
     });
-    const hasSolanaRows = payload.some((token) => {
-      return (
-        typeof token.chain === "string" &&
-        token.chain.toLowerCase() === ExternalChain.SOLANA
-      );
+    const scopedPayload = query.chain
+      ? visiblePayload.filter((token) => getTokenChain(token) === query.chain)
+      : visiblePayload;
+
+    const hasEthereumRows = scopedPayload.some((token) => {
+      return getTokenChain(token) === ExternalChain.ETHEREUM;
     });
-    const ethers = hasEthereumRows
-      ? await loadEthers("Bridge token parsing")
-      : undefined;
-    const solanaWeb3 = hasSolanaRows
-      ? await loadSolanaWeb3("Bridge token parsing")
-      : undefined;
+    const hasSolanaRows = scopedPayload.some((token) => {
+      return getTokenChain(token) === ExternalChain.SOLANA;
+    });
+    const unavailableChains = new Set<ExternalChain>();
+    let ethers: Awaited<ReturnType<typeof loadEthers>> | undefined;
+    let solanaWeb3: Awaited<ReturnType<typeof loadSolanaWeb3>> | undefined;
+
+    if (hasEthereumRows) {
+      if (query.chain === ExternalChain.ETHEREUM) {
+        ethers = await loadEthers("Bridge token parsing");
+      } else {
+        try {
+          ethers = await loadEthers("Bridge token parsing");
+        } catch (error) {
+          if (!isOptionalPeerDependencyError(error, "ethers")) {
+            throw error;
+          }
+          unavailableChains.add(ExternalChain.ETHEREUM);
+          console.warn(
+            '[starkzap] Skipping ethereum bridge tokens because optional peer dependency "ethers" is not installed.',
+            error
+          );
+        }
+      }
+    }
+
+    if (hasSolanaRows) {
+      if (query.chain === ExternalChain.SOLANA) {
+        solanaWeb3 = await loadSolanaWeb3("Bridge token parsing");
+      } else {
+        try {
+          solanaWeb3 = await loadSolanaWeb3("Bridge token parsing");
+        } catch (error) {
+          if (!isOptionalPeerDependencyError(error, "@solana/web3.js")) {
+            throw error;
+          }
+          unavailableChains.add(ExternalChain.SOLANA);
+          console.warn(
+            '[starkzap] Skipping solana bridge tokens because optional peer dependency "@solana/web3.js" is not installed.',
+            error
+          );
+        }
+      }
+    }
+
     const normalizeEthereumAddress = ethers
       ? (value: string) => fromEthereumAddress(value, ethers)
       : undefined;
@@ -332,9 +395,14 @@ export class BridgeTokenRepository {
       ? (value: string) => fromSolanaAddress(value, solanaWeb3)
       : undefined;
 
-    const tokens = payload
+    const tokens = scopedPayload
       .filter((token) => {
-        return !token.hidden && !token.deprecated;
+        if (isExplicitChainRequest) {
+          return true;
+        }
+
+        const chain = getTokenChain(token);
+        return chain === null || !unavailableChains.has(chain);
       })
       .map((token) => {
         try {
@@ -344,6 +412,9 @@ export class BridgeTokenRepository {
             normalizeSolanaAddress
           );
         } catch (e) {
+          if (isExplicitChainRequest) {
+            throw e;
+          }
           console.warn(`Ignoring token ${token.symbol} due to`, e);
           return null;
         }
--- a/tests/bridge-token-repository.test.ts
+++ b/tests/bridge-token-repository.test.ts
@@ -1,11 +1,17 @@
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { BRIDGE_TOKEN_CACHE_TTL_MS, BridgeTokenRepository } from "@/bridge";
+import * as ethersRuntime from "@/connect/ethersRuntime";
+import * as solanaWeb3Runtime from "@/connect/solanaWeb3Runtime";
 import {
   EthereumBridgeToken,
   ExternalChain,
   Protocol,
   SolanaBridgeToken,
 } from "@/types";
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
 
 function mockApiResponse() {
   return [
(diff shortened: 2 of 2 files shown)


0xLucqs (commented):
# PR #74 Review: Cross-Chain Bridging Support

## CRITICAL (5 issues — must fix before merge)

### 1. Missing `await` on `getAllowanceSpender()` in CCTP `to` field
`src/bridge/ethereum/cctp/CCTPBridge.ts:169-172`

`getAllowanceSpender()` returns `Promise<EthereumAddress>` but is assigned to `to` without `await`. The transaction would be sent to `[object Promise]`.

### 2. ERC20 approve race condition — no reset-to-zero
`src/bridge/ethereum/EthereumBridge.ts:123`

`approveSpendingOf` calls `approve(spender, amount)` directly. Tokens like USDT revert when changing a non-zero allowance to a different non-zero value. Must approve to 0 first, then to the desired amount.

### 3. CCTP `depositForBurn` — Starknet address not padded to `bytes32`
`src/bridge/ethereum/cctp/CCTPBridge.ts:156-167`

The `recipient` (Starknet felt252) can be shorter than 32 bytes. CCTP expects `bytes32 mintRecipient`. Must left-pad to 32 bytes before passing to ABI encoder.

### 4. Failed bridge promise cached permanently
`src/bridge/operator/BridgeOperator.ts:97-102`, `BridgeCache.ts:33-39`

The promise from `createBridge()` is cached before it resolves. If it rejects (RPC timeout, etc.), the rejected promise stays in cache forever. Every subsequent call returns the same failure. Need a `.catch` handler that evicts from cache on rejection.

### 5. Cache key missing `token.protocol`
`src/bridge/operator/BridgeCache.ts:42-44`

Key is `${network}:${address}:${token.address}`. Two tokens sharing the same L1 address but different protocols (e.g., canonical vs OFT_MIGRATED) would collide. Include `token.protocol` in the key.

---

## MAJOR (11 issues — strongly recommend fixing)

### 6. Missing `await` on `getFees()` in CCTPFees
`src/bridge/ethereum/cctp/CCTPFees.ts:44` — `this.getFees()` is async but not awaited. `Array.isArray(feeData)` is always `false` for a Promise, so live Circle API fees are **never used** — only hardcoded fallbacks.

### 7. `WalletInterface extends BridgeOperatorInterface` is a breaking change
`src/wallet/interface.ts:48` — Adds 4 required methods to the wallet contract. External implementors break at compile time. Consider a separate `BridgeableWallet` interface or accessor pattern (`wallet.bridge().deposit(...)`).

### 8. Missing Solana address runtime validation
`src/connect/solana.ts:68` — Ethereum path validates via `fromEthereumAddress()` with checksum. Solana path passes `options.address` straight through without calling `fromSolanaAddress()`.

### 9. `deposit()` silently succeeds with empty transaction list
`src/bridge/solana/SolanaHyperlaneBridge.ts:77-85` — If `getTransferRemoteTxs` returns `[]`, `lastSignature` stays `""` and method returns `{ hash: "" }`. Should throw.

### 10. No zero-amount validation on bridge deposits
`CanonicalEthereumBridge.ts:24`, `CCTPBridge.ts:43`, `OftBridge.ts:60` — None validate `amount > 0`. A zero deposit triggers approval flows and gas costs for nothing.

### 11. No bridge-specific error types
`src/types/errors.ts` — All bridge errors are bare `new Error(...)`. Consumers can't programmatically distinguish bridge misconfiguration from chain mismatch from unsupported protocol.

### 12. `BridgeOperatorInterface` not publicly exported
`src/bridge/operator/index.ts` — Only `BridgeOperator` is exported, but `WalletInterface` extends `BridgeOperatorInterface`. External consumers can't reference the interface.

### 13. OFT `dummyDepositTxPromise` cached forever
`src/bridge/ethereum/oft/OftBridge.ts:179-181` — Cached with `??=` and never invalidated. Gas price changes won't be reflected.

### 14. Relative import escapes module boundary
`src/wallet/base.ts:42` — `import type { ConnectedExternalWallet } from ".."` resolves to `src/index.ts`, creating circular dependency risk. Should be `from "@/connect"`.

### 15. `file:../..` dependency breaks npm publish
`packages/native/package.json:36` — `"starkzap": "file:../.."` will fail if published. Has `publishConfig.access: "public"`.

### 16. Stale `sideEffects` reference
`packages/native/package.json:26` — Points to `./dist/polyfills/install.js` which no longer exists after this PR deleted the polyfills directory.

---

## MINOR (14 issues)

| # | File | Issue |
|---|------|-------|
| 17 | `BridgeOperator.ts:116,122` | Loose equality (`==`) instead of `===` for chain comparison |
| 18 | `connect/evm.ts:117` | Error says "for Solana" but this is the Ethereum connector |
| 19 | `tokens/repository.ts:174` | Missing closing quote: `"${chain}` should be `"${chain}"` |
| 20 | `EtherToken.ts:167` | `"pending"` block tag unreliable across Ethereum nodes; prefer `"latest"` |
| 21 | Gas price logic | Duplicated with inconsistent implementations across Canonical, CCTP, OFT bridges |
| 22 | `CanonicalEthereumBridge.ts:113-118` | `recipient` passed without `.toString()`, inconsistent with other args |
| 23 | `solana/SolanaHyperlaneBridge.ts:139,171` | Swallowed errors in fee estimation — original error context lost |
| 24 | `solana/SolanaHyperlaneBridge.ts:30` | Floating-point rent: `BigInt(Math.round(0.00411336 * 1e9))` — use `BigInt(4_113_360)` directly |
| 25 | `connect/utils.ts:39-63` | `normalizeChainId` is dead code, duplicates logic already in `evm.ts` |
| 26 | `BridgeOperator.ts:142` | Hardcoded `"lords"` magic string — should be a named constant |
| 27 | `tokens/repository.ts:279` | `console.warn` in SDK library pollutes consumer output |
| 28 | `BaseWallet` bridge methods | No JSDoc, unlike every other public method |
| 29 | No bridge wallet tests | `deposit`/`getDepositBalance`/`getAllowance`/`getDepositFeeEstimate` delegation untested |
| 30 | `connect/index.ts:10-11` | `export *` leaks internal helpers like `assertEip1193Provider` |

---

## Summary

**Architecture**: Clean strategy pattern with `BridgeInterface` → `BridgeOperator` routing. Lazy loading for Solana is well done. Token repository with caching/dedup is solid.

**Must fix**: The CCTP issues (#1, #3, #6) could cause lost funds or silently wrong behavior. The cache bugs (#4, #5) will cause permanent failures after transient errors. The approve race condition (#2) will fail on USDT and similar tokens.

**Recommend fixing**: The breaking `WalletInterface` change (#7) is a semver concern worth discussing. Missing address validation (#8) and zero-amount checks (#10) are safety gaps.

**Test coverage**: Only `BridgeTokenRepository` has unit tests. The bridge delegation layer, individual bridge implementations, and wallet connection modules are untested.

0xLucqs (commented):
## Thorough Code Review -- Cross-Chain Bridging (PR #74)

Overall this is an ambitious and well-architected feature. The strategy pattern for bridge implementations, the operator/cache layering, and the wallet interface integration are all solid design decisions. That said, there are several issues that should be addressed before merging, organized by severity.

---

### Critical

**1. Solana address is never validated at runtime** (`src/connect/solana.ts:67`)

`ConnectedSolanaWallet.from()` passes `options.address` directly to the constructor without calling `fromSolanaAddress()`. The Ethereum path correctly validates via `fromEthereumAddress(options.address)` at `evm.ts:99`, but Solana skips this entirely. The `SolanaAddress` branded type is erased at runtime, so any arbitrary string is accepted. A malformed address would only surface during an actual bridge transaction.

**2. `CCTPFees`: No caching -- redundant API calls on every fee estimate** (`src/bridge/ethereum/cctp/CCTPFees.ts:42-43`)

The comment `// Possibly cache these fees.` acknowledges this. `getFees()` calls `fetchFees()` unconditionally, hitting the Circle API on every `getDepositFeeEstimate()` call. As a singleton shared across all `CCTPBridge` instances, every fee estimation triggers a network round-trip. This is a performance problem and a potential rate-limiting risk.

**3. `CCTPBridge.deposit()` and `OftBridge.deposit()`: Missing explicit gas limits** (`src/bridge/ethereum/cctp/CCTPBridge.ts:31-44`, `src/bridge/ethereum/oft/OftBridge.ts:56-71`)

Unlike `CanonicalEthereumBridge.deposit()` which calls `estimateEthereumSafeGasLimitForTx()`, both CCTP and OFT send transactions without explicit gas limits. The canonical bridge does this correctly -- the same pattern should be followed for consistency and safety.

---

### High

**4. `BridgeCache` has no size limit** (`src/bridge/operator/BridgeCache.ts`)

Both the ERC20 cache and the staking cache enforce limits (`MAX_ERC20_CACHE_SIZE = 128`, `MAX_STAKING_CACHE_SIZE = 128`) with LRU-style eviction. `BridgeCache` has no eviction at all. In long-running applications bridging many token/wallet combinations, this map grows without bound.

**5. `BridgeCache.key()` doesn't include token chain or protocol** (`src/bridge/operator/BridgeCache.ts:47`)

Key is `${wallet.network}:${wallet.address}:${token.address}`. If the same address appears on different chains (Ethereum vs Solana use different formats, but both are strings), or if a token migrates protocol (CANONICAL to OFT_MIGRATED), the cache returns the wrong bridge instance. Include `token.chain` and `token.protocol` (or `token.id`) in the key.

**6. `==` instead of `===` in `BridgeOperator.createBridge()`** (`src/bridge/operator/BridgeOperator.ts:100,105`)

```ts
if (token.chain == ExternalChain.ETHEREUM) {
```
Should be `===`. Both sides are strings, so no behavioral difference today, but this violates the project's strict conventions.

**7. `BridgeTokenRepository`: API response type is unchecked** (`src/bridge/tokens/repository.ts:188`)

```ts
const payload: BridgeTokenApiRecord[] = await response.json();
```
Casts with no runtime validation. If the API returns a different shape, this silently produces incorrect data or throws confusing errors downstream.

**8. `OftBridge.getAllowanceSpender()`: Failed fetch is cached forever** (`src/bridge/ethereum/oft/OftBridge.ts:87-104`)

`cachedSpender` is `null` after a failed fetch, and `null` is cached permanently with no retry. If the failure was transient (network blip), the OFT bridge permanently reports no spender, meaning allowance checks always return `null`.

**9. `LayerZeroApi` fetch binding** (`src/bridge/ethereum/oft/LayerZeroApi.ts:57`)

`private readonly fetcher: typeof fetch = fetch` -- unbound `fetch` can throw in some environments. `BridgeTokenRepository` correctly uses `globalThis.fetch.bind(globalThis)`. Same pattern should be used here.

**10. `OftBridge.dummyDepositTxPromise` cached forever** (`src/bridge/ethereum/oft/OftBridge.ts:37,122-124`)

This promise has no TTL. If LayerZero API returns stale fee data (since `msg.value` for interchain fees can change), the fee estimate will be inaccurate indefinitely.

---

### Medium

**11. Bridging methods on `BaseWallet` have no JSDoc** (`src/wallet/base.ts:755-791`)

Every other public method on `BaseWallet` has thorough JSDoc. These four new methods (`deposit`, `getDepositBalance`, `getAllowance`, `getDepositFeeEstimate`) are undocumented despite being part of the public API.

**12. `WalletInterface` inherits bridging silently** (`src/wallet/interface.ts:48`)

Bridging methods come from `BridgeOperatorInterface` via `extends`, but unlike ERC20 and staking sections (which have explicit declarations with comments), bridging methods are invisible when reading `interface.ts`. This breaks the established pattern.

**13. Cross-chain validation error messages are vague** (`src/connect/evm.ts:120-126`)

Ethereum errors say `"EVM chain id expected to be mainnet."` without mentioning the Starknet chain that caused the mismatch. The Solana equivalents (`solana.ts:61,65`) are better -- e.g., "Solana mainnet cannot be used with Starknet Sepolia". Ethereum messages should follow the same pattern.

**14. `BridgeTokenRepository`: Stale `now` in `fetchAndCache`** (`src/bridge/tokens/repository.ts:170`)

`now` is captured at the start of `getTokens()`, then used to compute `expiresAt` after the fetch completes. Should use `this.now()` at cache-write time.

**15. `SolanaHyperlaneBridge.deposit()`: No partial-failure handling** (`src/bridge/solana/SolanaHyperlaneBridge.ts:51-62`)

Signs and sends transactions sequentially. If the second fails, the first is already submitted. No mechanism to report which succeeded. Consider at minimum logging each transaction hash as it's submitted.

**16. `Eip1193Provider` interface includes non-standard parameters** (`src/connect/evm.ts:26-35`)

The `chain?` and `expiry?` parameters on `request()` are not part of EIP-1193. They appear to come from Reown/WalletConnect. This should be documented.

---

### Dependency Concerns

**17. Heavy new production dependencies**

`ethers` (~3.5MB), `@hyperlane-xyz/sdk`, `@hyperlane-xyz/registry`, and `@solana/web3.js` add significant bundle weight for **all** consumers, even those not using bridging. Consider:
- Making `ethers` and the Hyperlane/Solana packages optional peer dependencies (like `@cartridge/controller`)
- Lazy loading the Solana bridge already works via dynamic import -- the same approach could work for Ethereum bridges

---

### Test Coverage -- Major Gap

**18. Only 1 of 23 bridge source files has tests**

The PR adds 23 new files under `src/bridge/` + `src/connect/`, but only `BridgeTokenRepository` has unit tests. None of the bridge implementations, `BridgeOperator`, `BridgeCache`, `EtherToken`, `CCTPFees`, `LayerZeroApi`, or wallet connectors are tested.

**Most critical untested code:**
- `CCTPBridge.calculateMaxFee()` -- basis-point arithmetic touching user funds
- `EthereumBridge.approveSpendingOf()` -- allowance/approve flow
- `EthereumBridge.execute()` -- error mapping from ethers codes to SDK errors
- `BridgeOperator.createBridge()` -- factory dispatch and error paths
- `BridgeCache.set()` error self-eviction
- `ConnectedSolanaWallet.from()` / `ConnectedEthereumWallet.from()` -- validation logic

These can all be unit tested with mocked fetch/providers following the same pattern as `bridge-token-repository.test.ts`.

---

### Example Code

The web and mobile examples are well-structured and demonstrate the full bridge flow. The `BridgeController` class in the web example is a clean separation of concerns. A few notes:

- The mobile store (`stores/wallet.ts`) has grown to ~600 lines with bridge state mixed in. Consider extracting bridge state into a separate store or slice.
- Multiple `as` type assertions in the mobile store (e.g., `bridgeSelectedToken as EthereumBridgeToken`) could be avoided with discriminated union narrowing.

---

### Low-Severity (for completeness)

- `CanonicalEthereumBridge.estimateL1ToL2MessageFee` swallows all errors silently (no logging)
- `EthereumBridge.approveSpendingOf()` doesn't validate `amount` is positive
- `BridgeTokenRepository.cloneTokens` is a shallow copy (mitigated by `readonly` fields)
- `EthereumBridge.updateAllowanceFromReceipt()` has a TODO about parsing logs twice
- `isRecord` is exported from `connect/utils.ts` but only used internally (leaks to public API)
- `params` type in `Eip1193Provider.request` has redundant `| object` (already covered by `Record<string, unknown>`)

---

### Summary

| Severity | Count |
|----------|-------|
| Critical | 3 |
| High | 7 |
| Medium | 6 |
| Dependency | 1 |
| Test Coverage | 1 (major) |
| Low | 6 |

The architecture is sound and the feature is impressive in scope. The main blockers are: (1) the Solana address validation gap, (2) missing gas limits on CCTP/OFT deposits, (3) unbounded caching, and (4) the test coverage gap for financial calculation logic and error paths. The dependency weight question is worth discussing as a team decision.

Issue #80, opened 2026-03-13
Epic: StarkZap Execution Parity (SDK -> MCP -> CLI)


omarespejel:
PR #21 proved the model: thin MCP wrappers over StarkZap SDK calls with strong validation and write gating.
Now SDK surface has grown, and MCP/CLI are behind.

This epic closes that gap and makes StarkZap the canonical execution surface for both developers and agents.

## Current status (already shipped)
- MCP baseline shipped in #21: account, balance, transfer, execute (gated), deploy, staking, estimate-fee.
- Additional hardening/improvements landed in #61 (read-only address queries).
- This epic is parity expansion, not a fresh MCP start.

## Decision boundary
- `starkzap` owns execution (SDK + MCP + CLI).
- `starknet-agentic` owns policy/safety (session lifecycle, ERC-8004 identity/metadata, attestation/provenance, x402 policy/preflight).

### P0 (now)
- [ ] `starkzap_get_balances`
- [ ] `starkzap_get_quote`
- [ ] `starkzap_swap`
- [ ] `starkzap_build_swap_calls`
- [ ] `starkzap_build_calls`
- [ ] `starkzap_lending_markets`
- [ ] `starkzap_lending_deposit`
- [ ] `starkzap_lending_withdraw`
- [ ] `starkzap_lending_withdraw_max`
- [ ] `starkzap_lending_borrow`
- [ ] `starkzap_lending_repay`
- [ ] `starkzap_lending_position`
- [ ] `starkzap_lending_health`
- [ ] `starkzap_lending_quote_health`

### P1 (next)
- [ ] `starkzap_get_allowance`
- [ ] `starkzap_approve_exact`
- [ ] `starkzap_approve_if_needed`
- [ ] `starkzap_revoke_approval`
- [ ] Bridge tools once bridge SDK module is merged:
  - [ ] `starkzap_bridge_quote`
  - [ ] `starkzap_bridge_execute`
  - [ ] `starkzap_bridge_status`
  - [ ] `starkzap_bridge_routes`

## CLI parity (same core, no duplicate logic)
- [ ] Add `starkzap-cli` commands for all P0 MCP tools.
- [ ] CLI and MCP must call the same shared action handlers + schemas.

## Implementation rules
- No business logic in MCP/CLI adapters.
- Shared action layer: parse/validate -> SDK call -> normalized output.
- MCP tool names must use the `starkzap_` namespace prefix to avoid collisions with other MCP servers.
[...]

PR #81, opened 2026-03-14
feat(mcp): add swap + batched balances tools with hardening

omarespejel:
## Summary

Expands StarkZap MCP execution parity with swap and batch balance tools, plus security hardening from prior MCP review feedback patterns.

### Added MCP tools
- `starkzap_get_balances`
- `starkzap_get_quote`
- `starkzap_swap`
- `starkzap_build_swap_calls`

### Hardening included in this PR
- Enforce strict schemas for the new swap/get-balances inputs (reject unknown keys).
- Add token/provider input bounds and provider format validation.
- Add same-token swap protections:
  - schema-level guard for obvious same-token inputs
  - runtime guard after token resolution (covers symbol/address equivalence)
- Keep schema parity checks compatible with refined schemas (`ZodEffects` unwrap).

## Why

This is the first scoped step for issue #80 (SDK -> MCP -> CLI parity), while keeping scope reviewable and production-safe.

## Validation

Ran in `packages/mcp-server`:
- `npm run typecheck`
- `npm run test`

Result: `84/84` tests passing.

## Files

- `packages/mcp-server/src/core.ts`
- `packages/mcp-server/src/index.ts`
- `packages/mcp-server/tests/core.test.ts`
- `packages/mcp-server/tests/index.integration.test.ts`
- `packages/mcp-server/README.md`

Refs #80

### DIFF
--- a/packages/mcp-server/src/index.ts
+++ b/packages/mcp-server/src/index.ts
@@ -537,6 +537,10 @@ function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null;
 }
 
+const FELT252_UPPER_BOUND = BigInt(
+  "0x800000000000011000000000000000000000000000000000000000000000001"
+);
+
 type AmountMethod =
   | "toUnit"
   | "toFormatted"
@@ -692,9 +696,15 @@ function normalizeCallCalldataForResponse(
 ): string[] {
   const MAX_CALLDATA_ITEMS = 2048;
   const MAX_CALLDATA_ITEM_CHARS = 256;
-  const MAX_FELT_HEX_CHARS = 64;
   const CALLDATA_DECIMAL_REGEX = /^\d+$/;
   const CALLDATA_HEX_REGEX = /^0x[0-9a-fA-F]{1,64}$/;
+  const assertFelt252Value = (value: bigint, index: number): void => {
+    if (value >= FELT252_UPPER_BOUND) {
+      throw new Error(
+        `Invalid ${label} returned by SDK: calldata_${index} exceeds felt range.`
+      );
+    }
+  };
 
   if (!Array.isArray(calldata)) {
     throw new Error(
@@ -713,13 +723,8 @@ function normalizeCallCalldataForResponse(
           `Invalid ${label} returned by SDK: calldata_${index} must be non-negative.`
         );
       }
-      const hexValue = item.toString(16);
-      if (hexValue.length > MAX_FELT_HEX_CHARS) {
-        throw new Error(
-          `Invalid ${label} returned by SDK: calldata_${index} exceeds felt range.`
-        );
-      }
-      return `0x${hexValue}`;
+      assertFelt252Value(item, index);
+      return `0x${item.toString(16)}`;
     }
     if (typeof item === "number") {
       if (!Number.isSafeInteger(item) || item < 0) {
@@ -727,6 +732,7 @@ function normalizeCallCalldataForResponse(
           `Invalid ${label} returned by SDK: calldata_${index} must be a non-negative safe integer.`
         );
       }
+      assertFelt252Value(BigInt(item), index);
       return item.toString(10);
     }
     if (typeof item === "string") {
@@ -738,14 +744,11 @@ function normalizeCallCalldataForResponse(
       }
       if (CALLDATA_DECIMAL_REGEX.test(trimmed)) {
         const decimalValue = BigInt(trimmed);
-        if (decimalValue.toString(16).length > MAX_FELT_HEX_CHARS) {
-          throw new Error(
-            `Invalid ${label} returned by SDK: calldata_${index} exceeds felt range.`
-          );
-        }
+        assertFelt252Value(decimalValue, index);
         return trimmed;
       }
       if (CALLDATA_HEX_REGEX.test(trimmed)) {
+        assertFelt252Value(BigInt(trimmed), index);
         return trimmed;
       }
       throw new Error(
@@ -1492,6 +1495,8 @@ function buildToolErrorText(error: unknown): string {
     "Sponsored ",
     "Transaction ",
     "Address ",
+    "Swap ",
+    "Build swap calls:",
     "starkzap_",
   ];
   const hasSafePrefix = safeMessagePrefixes.some((prefix) =>
@@ -1782,6 +1787,19 @@ async function handleTool(
       assertDistinctSwapTokens(tokenIn, tokenOut, "Swap execution");
       const amountIn = parseAmountWithContext(parsed.amountIn, tokenIn, "swap");
       assertAmountWithinCap(amountIn, tokenIn, maxAmount);
+      const quote = await withTimeout("Swap quote precheck", () =>
+        wallet.getQuote({
+          tokenIn,
+          tokenOut,
+          amountIn,
+          ...(parsed.slippageBps !== undefined && {
+            slippageBps: BigInt(parsed.slippageBps),
+          }),
+          ...(parsed.provider !== undefined && { provider: parsed.provider }),
+        })
+      );
+      assertSwapQuoteShape(quote);
+      const quotedOutAmount = Amount.fromRaw(quote.amountOutBase, tokenOut);
       const feeMode: "sponsored" | undefined = parsed.sponsored
         ? "sponsored"
         : undefined;
@@ -1817,6 +1835,19 @@ async function handleTool(
         tokenOutAddress: tokenOut.address,
         amountIn: amountIn.toUnit(),
         amountInRaw: amountIn.toBase().toString(),
+        amountOut: quotedOutAmount.toUnit(),
+        amountOutRaw: quote.amountOutBase.toString(),
+        amountOutSource: "pretrade_quote",
+        ...(quote.routeCallCount !== undefined && {
+          routeCallCount: quote.routeCallCount,
+        }),
+        ...(quote.priceImpactBps !== undefined && {
+          priceImpactBps:
+            quote.priceImpactBps === null
+              ? null
+              : quote.priceImpactBps.toString(),
+        }),
+        ...(quote.provider !== undefined && { provider: quote.provider }),
       });
     }
 
--- a/packages/mcp-server/src/index.ts
+++ b/packages/mcp-server/src/index.ts
@@ -488,10 +488,17 @@ function withTransactionReference(
   error: unknown,
   txResult: { hash: string; explorerUrl?: string }
 ): Error {
-  const reason = error instanceof Error ? error.message : String(error);
-  return new Error(
+  const reason =
+    error instanceof Error
+      ? error.message
+      : isRecord(error) && typeof error.message === "string"
+        ? error.message
+        : summarizeError(error);
+  const wrapped = new Error(
     `${reason} Transaction hash: ${txResult.hash}. Reconcile on-chain state before retrying.`
   );
+  (wrapped as Error & { cause?: unknown }).cause = error;
+  return wrapped;
 }
 
 async function assertWalletAccountClassHashAfterSubmission(
--- a/packages/mcp-server/src/index.ts
+++ b/packages/mcp-server/src/index.ts
@@ -608,6 +608,7 @@ function assertSwapQuoteShape(quote: unknown): asserts quote is {
   priceImpactBps?: bigint | null;
   provider?: string;
 } {
+  const PROVIDER_ID_REGEX = /^[A-Za-z0-9._:-]+$/;
   if (!isRecord(quote)) {
     throw new Error(
       "Invalid swap quote returned by SDK: expected object shape."
@@ -644,9 +645,15 @@ function assertSwapQuoteShape(quote: unknown): asserts quote is {
       "Invalid swap quote returned by SDK: priceImpactBps must be bigint or null."
     );
   }
-  if (quote.provider !== undefined && typeof quote.provider !== "string") {
+  if (
+    quote.provider !== undefined &&
+    (typeof quote.provider !== "string" ||
+      quote.provider.length === 0 ||
+      quote.provider.length > 64 ||
+      !PROVIDER_ID_REGEX.test(quote.provider))
+  ) {
     throw new Error(
-      "Invalid swap quote returned by SDK: provider must be a string."
+      "Invalid swap quote returned by SDK: provider must be a safe provider id."
     );
   }
 }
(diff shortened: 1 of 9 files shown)


0xLucqs (commented):
## Code Review

### Issues

#### Major

**1. Unbounded concurrency in `starkzap_get_balances`**
Up to 32 tokens can be queried simultaneously via `Promise.all`, firing 32 parallel RPC calls. This could overwhelm rate-limited RPC endpoints. Consider adding a concurrency limit (e.g., `p-limit` or manual batching of 5-8 concurrent requests).

**2. `starkzap_swap` response omits output amount**
Returns `amountIn`/`amountInRaw` but no `amountOut`/`amountOutRaw`. Compare with `starkzap_get_quote` which returns both sides. This is a significant usability gap for AI agents that need to confirm what was received.

**3. `starkzap_build_swap_calls` semantic annotations**
Marked as `readOnlyHint: true` and `idempotentHint: true`, but it calls external swap aggregator APIs (e.g., AVNU) to build routes. While it doesn't execute on-chain, aggregator calls have side effects (rate limits, potential quote reservation) and quotes change over time, making `idempotentHint` questionable.

#### Minor

**4. Duplicated `tokenIn !== tokenOut` refinement**
The same refinement is defined twice: once in `swapInputSchema` and again in the `starkzap_swap` schema definition. Additionally, `assertDistinctSwapTokens` performs the same check at the resolved-address level. The Zod refinement only does case-insensitive string comparison, which won't catch cases where symbol and address refer to the same token (that's caught by `assertDistinctSwapTokens`). Consider whether the Zod-level refinement adds value.

**5. `Amount.fromRaw` in quote handler has no try/catch**
In the `starkzap_get_quote` handler, `Amount.fromRaw(quote.amountInBase, tokenIn)` is called without try/catch. If the SDK returns a value that causes it to throw, the error message might leak through the safe-prefix check.

**6. `STARKNET_ACCOUNT_ADDRESS` override not validated**
If a user provides an address that doesn't correspond to their private key, wallet operations will fail at transaction time with cryptic errors rather than at startup. Consider an early validation or at minimum a warning log.

**7. Missing `minLength: 1` on `tokenIn`/`tokenOut` in MCP `inputSchema`**
The Zod `tokenIdentifierSchema` has `.min(1)`, but the `inputSchema` for swap tools only specifies `maxLength: 128` without `minLength: 1`. The parity check doesn't catch this.

**8. `PROVIDER_ID_REGEX` duplicated**
Defined locally inside `assertSwapQuoteShape` instead of reusing the constant from `providerSchema` in core.ts.

**9. No duplicate token check in `starkzap_get_balances`**
Batch size is capped at 32 via Zod, but duplicate tokens in the batch are not rejected, which could waste RPC calls.

#### Nit

**10. `additionalProperties: false` inconsistently applied**
New swap/balances tools include it, but existing tools (`starkzap_transfer`, `starkzap_execute`) don't. Functionally fine since Zod schemas are the real validators, but worth noting.

**11. No test for `sponsored: true` through `starkzap_swap`'s Zod schema**
The `sponsored-preflight.integration.test.ts` only tests transfer, not swap.

---

### Positives

- Thorough multi-layer input validation (slippage bounds, provider format, token distinctness)
- Strong output validation (`assertSwapQuoteShape`, `normalizeCallForResponse`) — excellent defense-in-depth for a funds-handling system
- Good test coverage: schema validation, write-gating, malformed SDK responses, same-token rejection, felt range bounds
- Sponsored preflight hardening (pre-submission class hash verification) is a meaningful security improvement
- Schema parity system correctly updated for `ZodEffects` via `getObjectSchema()`
- New tools follow established patterns for rate limiting, concurrency policy, timeout wrapping, and error sanitization

### Recommendation

**Request changes** — primarily on #1 (unbounded concurrency) and #2 (missing swap output amount). The rest can be addressed in follow-ups.

packages/mcp-server/src/index.ts
```diff
@@ -1416,6 +1723,12 @@ async function handleTool(
       const feeMode: "sponsored" | undefined = parsed.sponsored
         ? "sponsored"
         : undefined;
+      if (feeMode === "sponsored") {
+        await assertWalletAccountClassHash(
```
0xLucqs: Sponsored transfer, execute, and swap all run assertWalletAccountClassHash before submission. On an undeployed account that throws before the paymaster can auto-deploy, so the first sponsored write is blocked even though the branch docs describe sponsored mode as auto-deploy capable. The new test file only covers the deployed class hash mismatch path, so this regression is currently untested.

omarespejel:
Live MCP E2E smoke completed on the PR head (`feat/mcp-parity-swap-balances`).

Validation performed with the built server (`packages/mcp-server/dist/index.js`) over MCP stdio on Starknet Sepolia:

- `tools/list` exposed the expected PR tools: `starkzap_get_balances`, `starkzap_get_quote`, `starkzap_build_swap_calls`, `starkzap_swap`, plus existing wallet tools.
- `starkzap_get_account` confirmed the account is deployed.
- `starkzap_get_balance` confirmed STRK/ETH balances.
- `starkzap_get_quote` succeeded for STRK -> ETH.
- `starkzap_build_swap_calls` returned the expected approve + AVNU route calls.
- Real `starkzap_swap` succeeded with MCP write gates enabled and a tight cap (`--enable-write --max-amount 0.001`).

Sepolia tx: https://sepolia.voyager.online/tx/0x5cb6935bc4deccdacee98733466fabf18710bb5157f6f3b8c4f5d2f13a2bbd

RPC receipt check: `ACCEPTED_ON_L2`, `SUCCEEDED`.

No unresolved review threads remain and upstream CI is green (`test` + `examples`).

omarespejel:
Checked the `0xLucqs` top-level review at https://github.com/keep-starknet-strange/starkzap/pull/81#pullrequestreview-3954401292 against the current PR head.

Addressed in the current branch:

- `starkzap_get_balances` now uses bounded concurrency (`mapWithConcurrencyLimit(..., 8, ...)`) instead of unbounded `Promise.all`.
- `starkzap_swap` returns `amountOut`, `amountOutRaw`, and `amountOutSource` from the pre-trade quote.
- `starkzap_build_swap_calls` is marked non-idempotent while remaining read-only (`idempotentHint: false`).
- duplicate token inputs are rejected before balance reads, including post-resolution duplicate token addresses.
- swap quote/response raw amount conversion is routed through guarded conversion instead of naked `Amount.fromRaw`.
- provider ID validation uses the shared `PROVIDER_ID_REGEX`.
- `STARKNET_ACCOUNT_ADDRESS` is validated and fails fast if the SDK connects a different account than the requested override.
- swap tool `inputSchema` has the `minLength`/`maxLength`/provider-pattern constraints covered by parity tests.
- sponsored swap coverage exists for undeployed preflight, normal execution response shape, and deployed class-hash mismatch.
- follow-up hardening commit `fdc9a90` deduplicates the swap same-token refinement helper and adds `additionalProperties: false` to transfer/execute JSON schemas and their nested call/transfer item objects.

Validation run locally:

- `npm run test --workspace starkzap-mcp -- core.test.ts index.integration.test.ts sponsored-preflight.integration.test.ts` -> 106 tests passed
- `npm run check --workspace starkzap-mcp` -> typecheck, tests, and build passed

Also rechecked review threads: PR #81 currently has no unresolved review threads.

omarespejel:
@0xLucqs ready for review! Thanks

0xLucqs:
## Senior review — findings (MAJOR + MINOR)

Reviewed `packages/mcp-server/src/core.ts`, `src/index.ts`, and the three test files under `tests/`. Overall solid, defensive work (dual-layer validation, schema parity, redaction, bounded batches). Below are the issues worth addressing before merge.

### MAJOR

1. **`STARK_CURVE_ORDER` is misnamed — it's the field prime, not the EC curve order.**
   `packages/mcp-server/src/core.ts:9-11` holds `0x0800000000000011…0001` = `2^251 + 17·2^192 + 1`, which is Starknet's FIELD_PRIME. The EC order is strictly smaller (starts `0x0800…0010FFFF…`). `privateKeySchema.refine()` at `core.ts:47` therefore accepts scalars in `[EC_order, FIELD_PRIME)` that are not valid EC scalars. Probability of hitting that window at random is ~2⁻⁶² (not a real attack), but the name and the user-facing error message ("less than Stark curve order") are both wrong and will mislead a future reader. Introduced in this PR (commit c964bd9).
   *Fix:* rename to `STARK_FIELD_PRIME` **and** replace the value with the real EC order (`0x0800…0010FFFFFFFFFFFFFFFFB781126DCAE7B2321E66A241ADC64D2F`), or at minimum fix the error-message wording.

2. **Fee mode is a boolean `sponsored?`, not a discriminated union.**
   `core.ts:335-368` models fees as `sponsored: z.boolean().optional()`, lifted to `"sponsored" | undefined` at `index.ts:1831, 1874, 1925`. The SDK just merged a proper `FeeMode` tagged union (commits 8dc9a60, c4ee643) with ERC20-gas-token support. The MCP surface erases that union — any future variant will be invisible from MCP until the schema is rewritten.
   *Fix:* replace with `feeMode: z.discriminatedUnion("kind", […])` mirroring the SDK.

3. **Error-cause is dropped in 4 sites despite the PR's "preserve post-check error cause" claim.**
   Only `withTransactionReference` at `index.ts:500` wires `{ cause }`. These don't:
   - SDK init catch, `index.ts:283-291`
   - Wallet init `.catch`, `index.ts:449-461`
   - `resolveReferenceStakingClassHash` + `resolvePoolTokenForOperation`, `index.ts:1061-1066` and `:1188-1195`
   - Swap-quote `Amount.fromRaw` wrap, `index.ts:786-798`
   *Fix:* `throw new Error(msg, { cause: error })` at each site.

4. **`FELT252_UPPER_BOUND` duplicates `STARK_CURVE_ORDER` across files.**
   `index.ts:610` holds the same bigint as `core.ts:9` under a different name and different claimed semantics. Two copies, two names, same value. Single-source this — ideally export `FIELD_PRIME` (or a `Felt` brand with a validator) from the SDK proper (`src/types/`), since it's a Starknet primitive, not an MCP concern.

5. **`schemaParityMismatches` does NOT verify `additionalProperties`.**
   The new JSON schemas declare `additionalProperties: false` and the zod counterparts are `.strict()`, but `assertSchemaParity` doesn't check this contract. A contributor dropping `.strict()` would not trip CI. PR #83 introduces the fix for this (`getObjectAdditionalProperties` / `getArrayItemObjectAdditionalProperties` in `core.ts`). Cherry-pick that check here.

6. **`isClassHashNotFoundError` substring fallback can silently neutralize sponsored preflight** for an actually-deployed account.
   `core.ts:1361-1408` (pre-existing, but this PR now relies on it). The explicit-code allowlist is good, but the substring fallback — requires only `"not found"` (or `"is not deployed"` / `"uninitialized contract"`) plus `"contract"` / `"class hash"` — can trigger on generic RPC-proxy 404 bodies whose text happens to contain "contract". That returns `{ deployed: false }`, so `assertWalletAccountClassHashIfDeployed` skips the class-hash preflight.
   *Fix:* allowlist exact RPC codes only; drop the substring branch or log WARN when it's taken.

### MINOR

- **No `.strict()` tests for `starkzap_swap` or `starkzap_build_swap_calls`.** Only `get_balances` / `get_quote` are negative-tested at `tests/core.test.ts:179-193`. Zod's `.extend().refine()` on a strict base preserves strictness — but that behavior is a subtle zod-internals invariant. Add two `safeParse` tests asserting `success === false` for `{ …valid, unexpected: true }`.

- **No server-side test for `MAX_SWAP_CALLDATA_CHARS` (65,536).** The adjacent `MAX_SWAP_CALLDATA_ITEMS` guard has a test at `index.integration.test.ts:815`; the char-count guard at `index.ts:2030-2041` does not.

- **Balances concurrency cap of 8 workers is not exercised.** `mapWithConcurrencyLimit` is the only thing standing between a caller and 32 parallel RPC balance calls. Happy-path test uses 2 tokens — worker count is never pinned.

- **Post-check `.cause` chain is never asserted.** Sponsored tests only `toContain()` the outer error text. A regression that drops `cause` wiring in `withTransactionReference` will ship green.

- **Built-server startup validation claim is unsubstantiated** by tests — no test spawns `dist/`.

- **`starkzap_get_balance.token` (single) keeps the old loose `z.string().min(1)`** at `core.ts:395-397`, while batched uses the hardened `tokenIdentifierSchema`. Reuse one schema.

- **Schema `.strict()` asymmetric across tools.** `transfer`, `execute`, staking, and `estimate_fee` still don't apply `.strict()` even though their JSON `inputSchema` declares `additionalProperties: false`. Zod silently strips unknown keys. Pre-existing; flagging for consistency given the PR framing.

- **`idempotentHint: true` on `starkzap_get_quote`** (`core.ts:655`) is debatable — quotes move with market/liquidity. Clients caching on that hint may serve stale amounts. Either add JSDoc explaining why it's `true`, or set it `false`.

- **`noUncheckedIndexedAccess` disabled in `packages/mcp-server/tsconfig.json`** diverges from the root project invariant. `mapWithConcurrencyLimit` passes `items[currentIndex]` (typed `T | undefined` under strict) to a `T` parameter without a guard; logic is sound but would not type-check under root conventions.

- **Stringly-typed error classification via `safeMessagePrefixes`** (`index.ts:1593-1615`). Every new feature has to register a prefix (this PR adds `"Swap "` and `"Build swap calls:"`). Fragile — a missed prefix turns a user-actionable error into an opaque `Operation failed. Reference: …`. A tagged `ToolError` class hierarchy would replace the allowlist with `instanceof` checks.

- **Redundant `as z.infer<typeof schemas.X>` cast pattern** at every handler entry — the dispatcher already returns a typed `args`. The cast is dead weight and actively hides that zod refinements (`.refine(hasDistinctSwapTokenInputs)`) don't narrow the inferred type, which is why `assertDistinctSwapTokens` / `assertUniqueResolvedTokens` have to run again. Consider typing the dispatcher generically in `name`.

- **`normalizeCallForResponse` will duplicate with PR #83.** Both PRs define effectively the same helper (this PR at `index.ts:871`, PR #83 at `index.ts:713`). If both land, one must consolidate. Promote `NormalizedCall` + its validator to `src/tx/` in the SDK.

- **Test isolation risk in `sponsored-preflight.integration.test.ts`.** `beforeAll` mutates `process.env`, `process.argv`, and `__STARKZAP_MCP_TESTING__`, then imports `src/index.js`. Module-cache state (e.g., `setNowProvider` patches at lines 138, 169) can't be unwound by `afterAll`. Commit history (multiple "fully restore process env" commits) confirms this is a rough edge. Consider `vi.resetModules()` or a separate vitest project.

- **`sanitizeExplorerUrl` silent-drop** on any `new URL` error (`index.ts:394-399`) — returns `undefined` with only a console log and no `requestId` correlation. Include the request id for operator debugging.

- **Sponsored-swap happy-path lives inside `sponsored-preflight.integration.test.ts:284-326`**, not the main integration file. That file's purpose is preflight; swap wiring should be tested alongside the other swap behaviors.

- **Tests use `setWalletSingleton(mock as unknown as Wallet)`** — they are unit tests wearing integration-test clothing. They would not catch real SDK return-shape drift.

---
I cross-verified the high-severity findings in the source; if anything here disagrees with your mental model, flag it and I'll re-check.

PR #83, opened 2026-03-14
feat(mcp): add hardened starkzap_build_calls tool

omarespejel:
## Summary
- add new read-only MCP tool: `starkzap_build_calls`
- wire tool to SDK tx builder path (`wallet.tx().add(...).calls()`) so MCP wraps SDK behavior instead of reimplementing
- enforce strict input validation/hardening for build-calls payload (bounded call count + strict nested object shape)
- add defensive output normalization for SDK-returned calls (address/entrypoint/calldata type + bounds checks)
- include `starkzap_build_calls` in read-only gating/rate-limit bucket and docs

## Hardening details
- strict zod schema for `starkzap_build_calls` with `.strict()` call objects
- MCP input schema includes `additionalProperties: false` at top-level and call item level
- deterministic response shape with calldata normalized to strings only
- explicit guard for SDK call-count mismatch (`expected N calls, received M`)
- safe error surface preserved (invalid payload errors remain actionable; sensitive context still redacted)

## Tests
- `npm run typecheck` (packages/mcp-server)
- `npm run test` (packages/mcp-server)
- `npm run build` (packages/mcp-server)

Added tests:
- schema bounds/strictness for `starkzap_build_calls`
- read-only exposure checks
- integration success path for tx-builder-backed call normalization
- integration failure paths for malformed SDK call payloads and count mismatches


0xLucqs (commented):
## Code Review

### Issues

#### Major

**1. Inconsistent `.strict()` / `additionalProperties` between `starkzap_build_calls` and `starkzap_execute`**
`starkzap_build_calls` enforces both `.strict()` on Zod and `additionalProperties: false` on JSON schema, but `starkzap_execute` (which accepts the same call shape) has neither. In a funds-handling system, this inconsistency should be resolved — either add `.strict()` to execute as well, or document why `build_calls` needs it but execute doesn't.

**2. Hardcoded tool name in parity checker**
The `additionalProperties` check for array items is special-cased to `starkzap_build_calls` (`if (name === "starkzap_build_calls" && fieldName === "calls")`). This breaks the generic philosophy of the parity checker and will silently rot if other tools later add `.strict()` to their call items. Should be generalized.

#### Minor

**3. `starkzap_build_calls` classified as read-only but calls `wallet.tx()`**
While `txBuilder.calls()` doesn't submit anything, the SDK's `tx()` method should be verified to not have side effects (e.g., nonce reservation). If it does, the read-only classification would be misleading.

**4. Missing sponsored execute preflight test**
`sponsored-preflight.integration.test.ts` only tests sponsored **transfer** preflight. The identical sponsored **execute** path has no test coverage. While they share `assertWalletAccountClassHashIfDeployed`, a test for execute would catch regressions if the paths diverge.

**5. `STARKNET_ACCOUNT_ADDRESS` override not validated at startup**
No runtime validation that the provided address is controlled by the private key. Misconfiguration would fail with cryptic signature errors at tx time. Consider a startup warning log or early class-hash check.

**6. Redundant `"Sponsored "` error prefix**
Added to the safe error prefix allowlist, but `"Sponsored transfer preflight"` / `"Sponsored execute preflight"` already match via `startsWith`. This also broadens the allowlist to any error starting with "Sponsored ".

**7. Unused intermediate variable**
`const builtCalls = builtCallsResponse` after the `Array.isArray` check serves no purpose — `builtCallsResponse` could be used directly.

#### Nit

**8. `callPath` naming inconsistency**
Error paths in `normalizeCallForResponse` use `calls_0_calldata_0` format (underscore-separated) while elsewhere the codebase uses bracket notation in Zod paths (`calls[0]`).

**9. Test uses unrealistic address `"0x2"`**
Valid felt but not a realistic Starknet address. Using a full-length address would match patterns used elsewhere in tests.

---

### Positives

- Clean `WalletDeploymentStatus` discriminated union pattern — eliminates accessing `deployedClassHash` when undeployed
- DRY refactoring: `getWalletDeploymentStatus()` and `assertWalletAccountClassHashIfDeployed()` replace duplicated try/catch blocks across multiple tools
- Thorough output normalization (`normalizeCallForResponse`, `normalizeCalldataItemForResponse`) handling string/bigint/number types with length + format validation
- Sponsored preflight is a meaningful security addition preventing mismatched class hash submissions to paymasters
- Good test coverage: happy path, malformed SDK responses (invalid calldata type), and call-count mismatches
- Schema parity checker extended with root-level `additionalProperties` checking

### Recommendation

**Request changes** — primarily on #1 (strict/additionalProperties inconsistency with `starkzap_execute`) and #2 (hardcoded tool name in parity checker). #4 (missing sponsored execute test) should also be addressed before merge.

packages/mcp-server/src/index.ts
```diff
@@ -1473,27 +1639,54 @@ async function handleTool(
...
+      const builtCalls = builtCallsResponse;
+      if (builtCalls.length !== requestedCalls.length) {
+        throw new Error(
+          `Invalid build calls response from SDK: expected ${requestedCalls.length} calls, received ${builtCalls.length}.`
+        );
+      }
+      const normalizedCalls = builtCalls.map((call, index) =>
+        normalizeCallForResponse(call, index)
+      );
+      return ok({
```
0xLucqs, line 1729: starkzap_build_calls returns normalized calls verbatim, but unlike build_swap_calls it does not cap aggregate calldata items or total serialized size. With the current schema bounds, one valid request can generate a multi-megabyte MCP response, which is likely to hit client or server message limits. The tests only cover a small happy path and malformed payloads, not this size boundary.

omarespejel:
Validation update before review:

- Addressed previous review feedback; no unresolved non-outdated review threads remain on this PR.
- Fixed the latest `starkzap_build_calls` size-bound review item by enforcing aggregate calldata and serialized response limits before SDK calls and after SDK response normalization.
- Fork mirror PR checks are green: `test`, `examples`, and PR-Agent on omarespejel/x#2.
- Local package check passed on the branch: `npm run check --workspace starkzap-mcp` (`typecheck`, `vitest`, `tsup build`) with 86 tests passing.
- Built MCP stdio smoke passed on Sepolia in read-only mode using `https://starknet-sepolia-rpc.publicnode.com`:
  - `tools/list` exposed `starkzap_build_calls` and `starkzap_estimate_fee`.
  - `starkzap_execute` stayed hidden without `--enable-execute`.
  - `starkzap_build_calls` returned a valid normalized response.
- Also fixed a built-server startup issue found during smoke testing by bundling local `starkzap` while keeping node_modules external.

I think this is ready for maintainer review.

omarespejel:
Live MCP E2E smoke completed on the PR head (`feat/mcp-build-calls-hardening`).

Validation performed with the built server (`packages/mcp-server/dist/index.js`) over MCP stdio on Starknet Sepolia:

- `tools/list` exposed the expected tools: `starkzap_build_calls`, `starkzap_estimate_fee`, `starkzap_get_account`, `starkzap_get_balance`, `starkzap_transfer`, etc.
- `starkzap_get_account` confirmed the account is deployed.
- `starkzap_get_balance` confirmed STRK/ETH balances.
- `starkzap_build_calls` was previously smoke-tested read-only against live Sepolia RPC.
- Real `starkzap_transfer` self-transfer succeeded with MCP write gates enabled and a tight cap (`--enable-write --max-amount 0.000001`).

Sepolia tx: https://sepolia.voyager.online/tx/0x55802f3d765feac9191f7fd6802207474a98fd4abc557e050ed04c4fe6dd406

RPC receipt check: `ACCEPTED_ON_L2`, `SUCCEEDED`.

No unresolved review threads remain and upstream CI is green (`test` + `examples`).

omarespejel:
✅ Checked the `0xLucqs` top-level review at https://github.com/keep-starknet-strange/starkzap/pull/83#pullrequestreview-3954406830 against the current PR head.

All listed items are handled:

1. ✅ `.strict()` / `additionalProperties` consistency: `starkzap_execute` and `starkzap_build_calls` now both enforce strict call-shape validation with matching `additionalProperties: false` JSON schemas.
2. ✅ Hardcoded parity checker: strictness parity checking is generalized for object schemas and nested array item object schemas instead of being hardcoded only to `starkzap_build_calls`.
3. ✅ Read-only `wallet.tx()` concern: `starkzap_build_calls` only builds unsigned calls; it does not submit or reserve nonce. The response path is guarded and remains read-only.
4. ✅ Sponsored execute preflight test: sponsored execute now has undeployed-account and class-hash mismatch regression coverage in `sponsored-preflight.integration.test.ts`.
5. ✅ `STARKNET_ACCOUNT_ADDRESS` override: behavior is documented and covered by runtime tests that fail if the SDK connects a different account than the requested override.
6. ✅ Redundant `Sponsored ` prefix: the safe error prefix allowlist remains narrow (`Sponsored transfer` / `Sponsored execute`), not a broad `Sponsored ` prefix.
7. ✅ Unused intermediate variable: removed; `builtCallsResponse` is guarded and used directly.
8. ✅ `callPath` naming: follow-up commit `21b35e4` changes SDK call error paths to bracket notation (`calls[0].calldata[0]`) and fixes the sanitizer so numeric bracket paths are not mistaken for bracketed IPv6 hosts.
9. ✅ Unrealistic test address: follow-up commit `21b35e4` replaces the `0x2` account override test address with full-length Starknet-style test addresses.

Also handled Lucas’s separate inline comment:

- ✅ Build-calls response size cap: `starkzap_build_calls` now caps aggregate calldata/response size; the inline thread is resolved and the comment has a 👍 reaction.

Validation run locally:

- ✅ `npm run test --workspace starkzap-mcp -- core.test.ts index.integration.test.ts sponsored-preflight.integration.test.ts` -> 86 tests passed
- ✅ `npm run check --workspace starkzap-mcp` -> typecheck, tests, and build passed

Review-thread state:

- ✅ PR #83 currently has `0` unresolved review threads.

omarespejel:
@0xLucqs ready for review! Thanks

0xLucqs:
## Senior review — findings (MAJOR + MINOR)

Reviewed `packages/mcp-server/src/core.ts`, `src/index.ts`, and the three test files under `tests/`. Structurally this is the tighter of the two MCP PRs in flight — smaller blast radius, stronger schema parity (adds `additionalProperties` checks), shared constants hoisted to module level, and `normalizeCallForResponse` uses the branded `Address`. No BLOCKERs. Items below.

### MAJOR

1. **`FELT_REGEX` and `CALLDATA_ITEM_REGEX` are shape-only, not felt252-range.**
   `packages/mcp-server/src/core.ts:6` accepts up to 64 hex chars (up to `2^256 − 1`), and `CALLDATA_ITEM_REGEX` at `core.ts:307` additionally allows up to 256 decimal chars. Neither compares `BigInt(value) < FIELD_PRIME`. The README at line 35 ("validated against Starknet felt252 format") overstates the guarantee — values above the Stark field prime (~62 hex chars) pass. Not a security hole (RPC will reject), but the PR claim isn't satisfied.
   *Fix:* add a `.refine(v => BigInt(v) < FIELD_PRIME)` to the schemas, or soften the README to "felt-shaped hex string". Ideally export a `Felt` brand + validator from the SDK's `src/types/` since this is a repo-wide Starknet primitive.

2. **Error-cause is dropped in several sites despite "error cause preserved through post-check".**
   The claim is true for the tool's own post-check path, but these throw sites drop `cause`:
   - SDK init catch, `index.ts:290-298`
   - Wallet init `.catch`, `index.ts:456-468`
   - `resolveReferenceStakingClassHash` + `resolvePoolTokenForOperation`, `index.ts:954-959` and `:1081-1088`
   *Fix:* `throw new Error(msg, { cause: error })` at each site.

3. **`isClassHashNotFoundError` substring fallback can silently neutralize sponsored preflight** for an actually-deployed account.
   `core.ts:1361-1408` (pre-existing, but this PR now relies on it for its preflight claim). The explicit-code allowlist is good; the substring fallback (`"not found"` + `"contract"` / `"class hash"`) can trigger on generic RPC-proxy 404 bodies whose text happens to contain "contract". That returns `{ deployed: false }`, so `assertWalletAccountClassHashIfDeployed` silently skips class-hash verification.
   *Fix:* allowlist exact RPC codes only; drop the substring branch or log WARN when it's taken.

### MINOR

- **SDK call-count mismatch guard is one-sided in tests.**
  `tests/index.integration.test.ts:555-577` only exercises `received < requested`. The opposite direction (SDK returns MORE calls than requested) is not covered, even though `index.ts:1717-1721` handles both sides. Swapping `!==` for `<` would ship green.

- **`assertBuildCallsBatchBounds` JSON.stringifies the call array twice.**
  Called pre-SDK at `index.ts:1703` and post-SDK at `:1725`. For maximum-sized batches this is ~5 MB serialized twice per request. Pre-SDK can be a cheap arithmetic upper bound (sum of calldata lengths × 260 + N × 200 chars); the real `JSON.stringify` size check belongs only on the SDK response, where calldata shapes may have changed (bigint → decimal growth).

- **Duplicated magic numbers across files.**
  - `entrypoint.length > 64` at `index.ts:734` duplicates `core.ts:313`.
  - `calldata.length > 2048` at `index.ts:749` duplicates `core.ts:329`.
  Export `MAX_ENTRYPOINT_CHARS = 64` and `MAX_CALLDATA_ITEMS_PER_CALL = 2048` from `core.ts` and reuse.

- **No test for `starkzap_build_calls` annotations.**
  PR #81 pins `readOnlyHint` / `idempotentHint` for `build_swap_calls` at `tests/core.test.ts:333-341`. This PR should mirror for `build_calls`. The schema-parity test won't catch a wrong annotation.

- **No negative test for an invalid `contractAddress` at `build_calls` entry.**
  The handler uses `validateAddressBatch` at `index.ts:1693-1697`. Aggregation behavior when one call item has a bad address (analogous to the pattern at `tests/core.test.ts:237-248`) is not covered.

- **No concurrent-call test for `build_calls`.**
  `runWithToolConcurrencyPolicy` is tested at `tests/index.integration.test.ts:744-765` with `starkzap_get_balance` + `starkzap_estimate_fee` but not `build_calls`. Worth a test that two concurrent requests don't share tx-builder state.

- **No negative test for the schema-parity check itself.**
  The check is exercised generically at `tests/core.test.ts:275-279`, but no test asserts "if someone removes `.strict()` from the zod schema OR removes `additionalProperties: false` from the MCP JSON schema, parity detection fails." Given the reliance on that contract, a tampered-definition test would make the guarantee explicit.

- **Pre-SDK serialized-size check uses the request shape, not the response shape** (`index.ts:780-787`).
  It's called on `requestedCalls` (pre-normalization) to bound a response metric. Fine because the post-SDK call also runs on `normalizedCalls`, but worth a one-line comment: `// pre-check rejects oversized input; post-check rejects SDK expansion`.

- **Test isolation risk in `sponsored-preflight.integration.test.ts`.**
  Same pattern as PR #81 — `beforeAll` mutates `process.env`, `process.argv`, and `__STARKZAP_MCP_TESTING__`, then imports `src/index.js`. `afterAll` can't unwind module-cache state. Consider `vi.resetModules()` or a separate vitest project. Less blast radius here since the surface is smaller.

- **Redundant `as z.infer<typeof schemas.X>` cast pattern** at every handler entry — the dispatcher already returns a typed `args`. Dead weight, and it hides the fact that `.strict()` produces a type identical to the base (i.e., strict-ness is a runtime property, not a TS narrowing). Your `normalizeCallForResponse` already does proper runtime re-validation, so this is mostly cosmetic — but worth cleaning up.

- **`normalizeCallForResponse` will collide with PR #81** (which defines an effectively identical helper at `index.ts:871`). Whichever PR lands second should import rather than redefine. Better: promote `NormalizedCall` + its validator to `src/tx/` in the SDK so neither PR owns it.

- **README "Available Tools" table** lists `starkzap_build_calls` *after* `starkzap_execute`. Since `build_calls` is the safer read-only primitive, listing it first (or annotating it as "read-only, preferred for inspection") steers LLM agents toward it.

- **Tool description at `core.ts:572-573`** advertises "Maximum 10 calls per build batch" but omits the 2048 total-calldata cap and the 262,144-serialized-char cap. MCP clients surface the description to the LLM; mentioning the tighter bound makes the error actionable without a round-trip.

- **`sanitizeExplorerUrl` silent-drop** on any `new URL` error (`index.ts:394-399`) — returns `undefined` with only a console log and no `requestId` correlation. Include the request id for operator debugging.

---
I cross-verified the high-severity findings in the source; if anything here disagrees with your mental model, flag it and I'll re-check.

PR #87, opened 2026-03-17
ft: add Troves support

EjembiEmmanuel:
## Summary

Adds a `Troves` module to StarkZap that integrates with the Troves HTTP API.
### New module: `src/troves/`

- **`troves.ts`** – `TrovesStrategies` class:
  - Base URL: `https://app.troves.fi`
  - Constructor: `new TrovesStrategies(wallet, { fetcher?, timeoutMs? })`
  - Read: `getStrategies({ noCache? })`, `getStats()`
  - Write: `populateDepositCalls()`, `populateWithdrawCalls()`, `deposit()`, `withdraw()`
  - Normalizes API calldata to Starknet `Call[]`

- **`types.ts`** – Types for Troves API responses:
  - `TrovesStrategyAPIResult`, `TrovesStrategiesResponse`, `TrovesStatsResponse`
  - `TrovesDepositCallsResponse`, `TrovesRawCall`, `TrovesCallResult`, etc.

- **`index.ts`** – Re-exports for the Troves module

### Tests

- **`tests/troves.test.ts`** – 9 unit tests with mocked fetcher and wallet for:
  - `getStrategies` (with and without `noCache`)
  - `getStats`
  - `populateDepositCalls` / `populateWithdrawCalls`
  - `deposit` / `withdraw`
  - Error handling (API failures, empty results)

### Other

- **`src/index.ts`** – Added `export * from "@/troves"`

## API endpoints used

| Endpoint | Method | Purpose |
|----------|--------|---------|
| `/api/strategies` | GET | List strategies (APY, TVL, metadata) |
| `/api/stats` | GET | Aggregated TVL |
| `/api/deposits/calls` | POST | Build deposit/withdraw transaction calls |

## Usage

import { StarkZap, Troves } from "starkzap";

const sdk = new StarkZap({ network: "mainnet" });
const wallet = await sdk.connectWallet({ account: { signer } });
const troves = new Troves(wallet);

// Read
const strategies = await troves.getStrategies();
const stats = await troves.getStats();

// Write
const tx = await troves.deposit(
  { strategyId: "evergreen_strk", amountRaw: "1000000000000000000" },
  {}
);


src/troves/troves.ts
```diff
@@ -0,0 +1,259 @@
...
+import { fromAddress, type ExecuteOptions } from "@/types";
+import type { Tx } from "@/tx";
+import type {
+  TrovesStrategiesResponse,
+  TrovesStatsResponse,
+  TrovesDepositCallsResponse,
+  TrovesRawCall,
+} from "@/troves/types";
+
+const TROVES_API_BASE = "https://app.troves.fi";
```
0xLucqs, line 13: This client hardcodes one Troves environment and never checks the wallet chain before fetching strategies or building calls. On an unsupported Starknet network it will still return calls for app.troves.fi instead of failing fast. The lending providers already solve this with explicit chain support checks, and this module needs the same guard or an explicit environment option.

src/troves/troves.ts
```diff
@@ -0,0 +1,259 @@
...
+  private readonly wallet: WalletInterface;
+  private readonly fetcher: typeof fetch;
+  private readonly timeoutMs: number;
+
+  constructor(wallet: WalletInterface, options?: TrovesOptions) {
+    this.wallet = wallet;
+    this.fetcher =
+      options?.fetcher ??
+      ((url: RequestInfo | URL, init?: RequestInit) => fetch(url, init));
+    this.timeoutMs = options?.timeoutMs ?? 15000;
```
0xLucqs, line 123: timeoutMs is public but never validated. Zero, negative, or non-finite values all reach setTimeout directly, which can abort every request immediately. Other timeout options in this repo validate their bounds, and this one should too with a regression test.

0xLucqs (changes requested):
Two concrete issues block this for me: the missing chain guard and the unvalidated timeoutMs. Directionally, this also feels closer to the start of a shared earn surface than a Troves only helper. I would treat that earn interface work as follow up rather than requiring it in this PR, but I would avoid hardening the current wallet bound raw string shape more than necessary.

0xLucqs (changes requested):
## Review

Nice work getting the Troves module wired up — the overall structure (HTTP calls, address normalization via `fromAddress`, calldata conversion to starknet `Call[]`, and delegation to `wallet.execute()`) is solid. A few things to address before merging:

---

### Request Changes

**1. Extract duplicated `populateDepositCalls` / `populateWithdrawCalls`**

These two methods are ~38 lines each and share identical logic for address resolution, body construction, fetch, normalization, flattening, and both error guards. The only differences are `isDeposit: true` vs `false` and the word `"deposit"` vs `"withdraw"` in two error strings.

Extract a private method:

```ts
private async populateCalls(params: TrovesCallParams, isDeposit: boolean): Promise<Call[]>
```

This also eliminates the inline param type `{ strategyId: string; amountRaw: string; amount2Raw?: string; address?: string }` which is currently repeated across both populate methods, and the narrower version in `deposit` / `withdraw`. Define a named type in `types.ts`.

**2. `normalizeCalldata`: remove `(v: unknown)` widening and dead branches**

`TrovesRawCall.calldata` is typed `(string | number | boolean)[]`. The explicit `(v: unknown)` annotation overrides this, creating branches the compiler knows are unreachable:

- `typeof v === "bigint"` — dead. `bigint` is not in the union, and `JSON.parse` never produces bigints.
- `typeof v === "object"` — dead. `object` is not in the union.
- `Array.isArray(raw.calldata)` — always `true` per the declared type.

The fix: remove `(v: unknown)`, drop the bigint/object branches, and drop the `Array.isArray` guard. The compiler enforces the declared type. If the API genuinely can return values outside `(string | number | boolean)[]`, widen the type in `types.ts` to match reality — don't escape the type system in the consumer.

**3. Same pattern in `validateStrategiesDiscontinuationDates`**

After the `raw === undefined || raw === null` guard, `raw` is narrowed to `string` (the type is `date?: string`). So `typeof raw !== "string"` is a dead branch from the compiler's perspective. Additionally, `raw === null` itself is unnecessary since the type is `string | undefined`, never `null`.

If this is meant as defensive runtime validation against untrusted API data (which is reasonable at a boundary), then the type should reflect what the API might actually return — e.g. `date?: string | null` — so the compiler and the runtime checks agree rather than contradict each other.

**4. Tests: `wallet as never` and `{ hash } as Tx` bypass type checking**

Every test uses `new Troves(wallet as never, ...)`. The mock wallet only has `{ address, execute }` while `WalletInterface` has ~20 methods. `never` is assignable to any type, so the compiler skips all validation.

Similarly, `execute.mockResolvedValue({ hash: "0xmocktxhash" } as Tx)` creates a plain object cast to the `Tx` class, which has `wait()`, `watch()`, `receipt()` etc. The mock doesn't have any of them.

If `WalletInterface` or `Tx` change shape, these tests won't catch the mismatch at compile time. Use `Pick<WalletInterface, 'address' | 'execute'>` for the mock type and constrain accordingly, rather than casting to `never`.

---

### Suggestions (non-blocking)

**5. Use `resolveFetch()` from `src/bridge/utils/resolve-fetch.ts`**

The constructor's closure `(url, init) => fetch(url, init)` works, but `resolveFetch()` handles the `globalThis` binding and throws a descriptive error when `fetch` is unavailable (vs a raw `ReferenceError`). Reusing the existing utility is more consistent with the codebase.

**6. Wrap timeout `AbortError` with context**

When the timeout fires, users get a raw `DOMException: The operation was aborted` with no indication of which API path or what the timeout was. The staking module wraps these: `"Staking.fromPool timed out while {label}"`. Consider catching `AbortError` in `fetchJson` and rethrowing with `"Troves API request to ${path} timed out after ${this.timeoutMs}ms"`.

**7. Missing test coverage**

- No test exercises timeout/abort behavior despite `timeoutMs` being configurable.
- `validateStrategiesDiscontinuationDates` has 3 code paths (skip undefined, throw on bad type, throw on invalid date) — none are tested since the mock data has no `discontinuationInfo`.
- `amount2Raw` and `address` override params are accepted by both populate methods but never tested.

**8. Lock file drift**

The `package-lock.json` diff (149 additions, 796 deletions) is mostly removing `"peer": true` from babel packages and adding it to zod — unrelated to Troves. Should be reverted or split into a separate commit.

---

### Architecture direction (FYI, no action needed now)

The lending module uses a provider pattern: `LendingProvider` interface → `VesuLendingProvider` implementation → `LendingClient` dispatcher with `registerProvider()` → `wallet.lending()` accessor. When we add more yield/earn sources, we'll likely want a similar `EarnProvider` / `EarnClient` / `wallet.earn()` structure. No need to build that now, but structuring the Troves-specific logic (API endpoints, response shapes, calldata normalization) so it's easy to pull into a provider implementation later will make that migration straightforward. The `populateCalls` extraction from point 1 already moves in that direction.

0xLucqs:
@EjembiEmmanuel will you have time to finish that ? if not i can finish it just lmk

NeelkanthTandel:
> @EjembiEmmanuel will you have time to finish that? If not, I can finish it, just lmk

@0xLucqs, by when do you want?
We can try to address the review comments by the end of this week.

0xLucqs:
we want to release on thursday so ideally by tomorrow

0xLucqs:
@NeelkanthTandel Do you have an endpoint to get user positions or you're directly querying contracts through rpc ?

PR #92, opened 2026-03-17
fix: trim and validate network string input

owizdom:
## Problem

Network strings with trailing whitespace (e.g. `"sepolia\n"` from environment variables) cause a misleading error:

```
StarkZap requires either 'network' or 'rpcUrl' to be specified
```

The network **was** specified — it just has a trailing newline. `networks["sepolia\n"]` returns `undefined`, and the error then claims network wasn't provided at all.

This is common when env vars are copy-pasted from dashboards (Vercel, Railway) that sometimes include trailing whitespace.

## Fix

1. `.trim()` the network string before lookup
2. If the trimmed string doesn't match a valid network, throw a descriptive error listing valid options:
   ```
   Invalid network "sepolia\n". Valid options: mainnet, sepolia, devnet
   ```

## Context

Found while building [Zapp](https://github.com/owizdom/zap) (28 SDK modules used). Lost hours debugging this before spotting raw `\n` bytes in a Vercel env var.

Closes #89

### DIFF
--- a/src/sdk.ts
+++ b/src/sdk.ts
@@ -1,7 +1,7 @@
 import { RpcProvider, type Call, type PaymasterTimeBounds } from "starknet";
 import { ChainId, getChainId, type SDKConfig } from "@/types/config";
 import type { ConnectWalletOptions, FeeMode } from "@/types/wallet";
-import { networks, type NetworkPreset } from "@/network";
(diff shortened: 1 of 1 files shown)


0xLucqs (changes requested):
coderabbit suggested something that makes sense imo

0xLucqs:
hello hello still interested ?

PR #115, opened 2026-03-31
feat: bridge withdraw + monitoring

micbakos:
## Summary

- Implements initiate/complete withdraw flow for Canonical, CCTP, OFT, and Solana Hyperlane bridge protocols
- Adds auto-withdraw fee estimation and handling for canonical Ethereum bridges
- Implements full bridge monitoring layer (`BridgeMonitorInterface`) for all protocols with deposit and withdrawal state tracking:
  - **Canonical**: Ethereum receipt + Starknet event scanning
  - **CCTP**: Circle Iris API for attestation + Starknet `MessageReceived` event lookup by nonce
  - **OFT / OFT_MIGRATED**: LayerZero Scan API (`/messages/tx/{txHash}`) for both deposit and withdrawal — unified across both protocols, no API key required for monitoring
  - **Solana Hyperlane**: Hyperlane mailbox delivery checks in both directions
- Fixes pre-existing unit test failures: `auto-withdraw-fees-handler`, `EthereumBridge.approveSpendingOf`, `BridgeTokenRepository` AW_support field


0xLucqs (commented):
## Review — library only

### 🔴 `getAutoWithdrawTransferCall` crashes when wallet has no affordable gas token

`src/bridge/ethereum/canonical/CanonicalEthereumBridge.ts:336-337`

```ts
const feeTokenAddress = preferredEntry?.[0].address ?? feeData.preselectedGasToken!.tokenAddress;
const gasCost = preferredEntry?.[1] ?? feeData.preselectedGasToken!.cost;
```

`AutoWithdrawFeeOutput.preselectedGasToken` is typed as `{ ... } | undefined` and returns `undefined` when no wallet token can afford the gas cost. `getInitiateWithdrawFeeEstimate` correctly guards against this (throwing a descriptive error), but `initiateWithdraw` calls `getAutoWithdrawTransferCall` directly with no prior check — so the `!` assertions will throw an opaque `TypeError` at runtime.

Fix: add the same guard before calling `getAutoWithdrawTransferCall` in `initiateWithdraw`:
```ts
if (!feeData.preselectedGasToken) {
  throw new Error("The user has no sufficient balance to cover for auto-withdraw.");
}
```

---

### 🟡 `fetchAutoWithdrawData` missing `AbortSignal` timeout

`src/bridge/utils/auto-withdraw-fees-handler.ts:110`

```ts
const response = await this.fetchFn(url.toString(), {
  method: "GET",
  headers: { Accept: "application/json" },
});
```

Every other external HTTP call in this PR uses `signal: AbortSignal.timeout(10_000)`. A slow or hung SpaceShard endpoint will cause both `getInitiateWithdrawFeeEstimate` and `initiateWithdraw` (on the auto-withdraw path) to hang indefinitely.

---

### 💬 nit: `CompleteBridgeWithdrawOptions` is CCTP-only but used across all protocols

`src/bridge/types/BridgeInterface.ts:113`

```ts
export type CompleteBridgeWithdrawOptions = ExecuteOptions & CCTPCompleteBridgeWithdrawOptions;
```

`attestation: string` and `message: string` are required fields, but `EthereumBridge.completeWithdraw` (canonical) ignores `options` entirely (`_options?`). No runtime impact, but the interface implies requirements that don't apply to all implementors. Could be a union mirroring `InitiateBridgeWithdrawOptions` to be more honest about what each protocol actually needs.

0xLucqs (commented):
test comment to verify auth

0xLucqs (commented):
## Review Summary

Architecture is solid — consistent monitor interface, lazy dynamic imports, DI for testability, uniform `AbortSignal.timeout(10_000)` on most fetches. A few things worth addressing before or shortly after merge:

---

### 🔴 Blocker

**`CCTPBridge.waitForReattestation` uses bare global `fetch`**
`src/bridge/ethereum/cctp/CCTPBridge.ts:456,470` — every other fetch in this PR goes through `resolveFetch`, but this function calls raw `fetch`. Will throw `ReferenceError: fetch is not defined` in Node < 18 / React Native / workers when a CCTP withdrawal needs re-attestation. Trivial fix — wire through `resolveFetch` like the rest of the codebase.

---

### 🟠 Should fix

**`CCTPFees.fetchFees` also missing `AbortSignal.timeout`**
`src/bridge/ethereum/cctp/CCTPFees.ts:69` — a hung Circle API will hang the entire deposit/withdraw fee path forever. Same issue in `LayerZeroApi.getQuotes` (`src/bridge/ethereum/oft/LayerZeroApi.ts:179`). The recent "add abort timeouts" commit missed these two call sites.

**`CCTPBridge.getCompleteWithdrawFeeEstimate` inner catch returns fallback gas with no error flag**
`src/bridge/ethereum/cctp/CCTPBridge.ts:284-290` — when `estimateGas` fails, it falls through to `FALLBACK_COMPLETE_WITHDRAW_GAS * gasPrice` with no `l1FeeError` set. The outer catch at `:291-298` does set the flag, but this inner path silently returns a fabricated number that callers will trust as a real estimate. Set `l1FeeError` on the fallback path too.

**`snTxHash` vs `starknetTxHash` JSDoc mismatch**
`BridgeOperatorInterface.ts:230` and `BridgeMonitorInterface.ts:54` both document the accepted shape as `{ snTxHash, externalTxHash? }`, but the actual `WithdrawalStateInput` type uses `starknetTxHash`. Anyone copying from the docs gets a compile error.

**`CctpWithdrawMonitorResult` JSDoc makes a false guarantee**
`src/bridge/monitor/types.ts:42-47` — says "When `status` is `COMPLETED_ON_STARKNET`, the Circle attestation fields are populated and can be passed directly to `completeWithdraw`." In practice `CctpMonitor.monitorWithdrawal` can return `COMPLETED_ON_STARKNET` with `attestation: undefined` (Starknet finalized but Circle hasn't attested yet). Callers trusting the doc and feeding the result into `completeWithdraw` will crash. Either fix the doc or make the runtime match it.

**Dead exported constants**
`src/bridge/ethereum/cctp/constants.ts:82-83` — `ATTESTATION_POLL_INTERVAL_MS` and `ATTESTATION_MAX_POLL_ATTEMPTS` are exported but never imported anywhere. Leftover from an earlier design.

**`resolveMonitor` race condition**
`src/bridge/operator/BridgeOperator.ts:257-331` — uses `if (!this._monitors.has(p)) { await import; set }`. Two concurrent callers both pass the check, both construct monitors, second overwrites first. The first caller uses an orphaned instance. Fix with an in-flight promise map (same pattern used for `OftBridge.dummyDepositTxCache` and `CctpMonitor.getMessageTransmitter` elsewhere in this PR).

---

### 🟡 Worth noting (non-blocking)

- **Monitor layer silent failures**: every `tryFetch*` / delivery-check / event-scan in `CctpMonitor`, `OftMonitor`, `SolanaHyperlaneMonitor` collapses HTTP/RPC/parse errors into `null` or `NOT_SUBMITTED`, indistinguishable from "data not yet available". Works fine on the happy path, but during an upstream API outage users will see "pending" forever instead of an error indicator. Consider a discriminated result type (`ok | notFound | error`) in a follow-up.

- **Fee estimators return `Amount.fromRaw(0n, ...)` as fallback** across ~10 call sites with an optional `*FeeError` companion field. Downstream consumers that add the numbers without checking the error flag will present "free" to users. A follow-up could make the fee field `undefined` on the error path so ignoring the error is a type error.

- **`InitiateBridgeWithdrawOptions`** (`BridgeInterface.ts:71-72`) only covers `"canonical" | "cctp"` — OFT/Hyperlane callers wanting to pass `feeMode` or `timeBounds` are forced to satisfy one of those discriminants. Not blocking (options can be omitted), but worth a protocol-agnostic arm.

- **`BridgeTransferStatus.CONFIRMED_ON_L1`** is reused by `SolanaHyperlaneMonitor` to mean "delivered to Solana" (with an apology comment at `:131`). Works through `getWithdrawalState`, but any direct consumer switching on the enum will misinterpret Hyperlane results.

- **`BaseWallet.monitorDeposit/monitorWithdrawal/getDepositState/getWithdrawalState`** have zero JSDoc while every other method on the class has full `@param`/`@returns`/`@throws`/`@example`. These are the PR's headline feature — worth a doc pass.

- **Zero test coverage for `src/bridge/monitor/**`** (~1,300 LOC). The `auto-withdraw-fees-handler` test is great; the monitoring layer would benefit from the same treatment, especially CCTP state machine transitions and `deriveStarknetDepositTxHash` (a pure function that's trivially fixture-testable).

0xLucqs (approved):

PR #119, opened 2026-04-03
feat: add gasToken option to execute and deploy

### LINKED ISSUE #118: Add gasToken option to execute and deploy
## Problem

The SDK currently supports two fee modes: `"user_pays"` (pay in ETH) and `"sponsored"` (dApp pays). There is no way for a user to pay gas fees using an ERC-20 token like USDC or STRK.

However, `starknet.js` already supports this via the AVNU Paymaster:

```typescript
await account.executePaymasterTransaction(calls, {
  feeMode: { mode: 'default', gasToken: USDC_ADDRESS },
});
```

StarkZap itself re-exports this type as `PaymasterFeeMode` in `src/types/sponsorship.ts`, but it's not exposed in the wallet API.

## Current workaround

Without SDK support, applications must monkey-patch the internal account method:

```typescript
const account = wallet.getAccount() as any;
const original = account.executePaymasterTransaction.bind(account);
account.executePaymasterTransaction = async (calls, details, ...rest) =>
  original(calls, { ...details, feeMode: { mode: 'default', gasToken: USDC } }, ...rest);
```

This bypasses type safety, depends on SDK internals, and can break on updates.

## Proposal

Add an optional `gasToken` field to `ExecuteOptions` and `DeployOptions`:

```typescript
export interface ExecuteOptions {
  feeMode?: FeeMode;
  timeBounds?: PaymasterTimeBounds;
  gasToken?: Address; // new
}
```

When `gasToken` is provided, use `{ mode: 'default', gasToken }` via the paymaster. When omitted, behavior is unchanged.

```typescript
await wallet.execute(calls, { gasToken: USDC_ADDRESS });
```

ahmetenesdur:
Closes #118

## What

Adds an optional `gasToken` field to `ExecuteOptions` and `DeployOptions`, allowing users to pay gas fees with ERC-20 tokens (USDC, STRK, etc.) through the AVNU Paymaster.

When `gasToken` is set, the SDK calls `executePaymasterTransaction` with `{ mode: 'default', gasToken }` instead of `{ mode: 'sponsored' }`. When omitted, nothing changes.

## Changes

| File | What changed |
|------|-------------|
| `src/types/wallet.ts` | Added `gasToken?: Address` to `ExecuteOptions` and `DeployOptions` |
| `src/wallet/utils.ts` | `sponsoredDetails()` now accepts `gasToken` and produces the correct fee mode |
| `src/wallet/index.ts` | `execute()` and `deploy()` read and forward `gasToken` through the full chain |
| `src/wallet/cartridge.ts` | `execute()` forwards `gasToken`; `deploy()` guards against it |

## Usage

```typescript
// Pay gas with USDC
await wallet.execute(calls, {
  gasToken: "0x053c91253bc...368a8",
});

// Sponsored (unchanged)
await wallet.execute(calls, { feeMode: "sponsored" });

// User pays ETH (unchanged)
await wallet.execute(calls);
```

## Tests

All 116 unit tests pass. No breaking changes.

### DIFF
--- a/src/types/wallet.ts
+++ b/src/types/wallet.ts
@@ -223,6 +223,13 @@ export interface PreflightOptions {
    * because the paymaster path can deploy + execute atomically.
    */
   feeMode?: FeeMode;
+  /**
+   * ERC-20 token contract address used to pay gas fees via the paymaster.
+   *
+   * When set and the account is undeployed, preflight returns `{ ok: true }`
+   * because the paymaster path handles deployment atomically.
+   */
+  gasToken?: Address;
 }
 
 /** Preflight succeeded — operation can proceed */
--- a/src/types/wallet.ts
+++ b/src/types/wallet.ts
@@ -74,7 +74,12 @@ export interface AccountConfig {
 export type FeeMode =
   | "user_pays"
   | { type: "paymaster"; gasToken?: Address }
-  | "sponsored";
+  | DeprecatedSponsoredFeeMode;
+
+/**
+ * @deprecated Use `{ type: "paymaster" }` instead.
+ */
+type DeprecatedSponsoredFeeMode = "sponsored";
 
 // ─── Provider Options ────────────────────────────────────────────────────────
 
--- a/src/types/wallet.ts
+++ b/src/types/wallet.ts
@@ -65,10 +65,16 @@ export interface AccountConfig {
 
 /**
  * How transaction fees are paid.
- * - `"sponsored"`: Paymaster covers gas
- * - `"user_pays"`: User's account pays gas in ETH/STRK
+ *
+ * - `"user_pays"` — User's account pays gas in ETH/STRK
+ * - `{ type: "paymaster" }` — Paymaster covers gas (sponsored)
+ * - `{ type: "paymaster", gasToken: "0x..." }` — Pay gas via ERC-20 through paymaster
+ * - `"sponsored"` — *(deprecated)* Alias for `{ type: "paymaster" }`
  */
-export type FeeMode = "sponsored" | "user_pays";
+export type FeeMode =
+  | "user_pays"
+  | { type: "paymaster"; gasToken?: Address }
+  | "sponsored";
 
 // ─── Provider Options ────────────────────────────────────────────────────────
 
@@ -109,7 +115,7 @@ export interface ProviderOptions {
  * // Sponsored via AVNU paymaster
  * await sdk.connectWallet({
  *   account: { signer: new StarkSigner(privateKey) },
- *   feeMode: "sponsored",
+ *   feeMode: { type: "paymaster" },
  * });
  * ```
  */
@@ -154,7 +160,7 @@ export interface ProgressEvent {
  * ```ts
  * await wallet.ensureReady({
  *   deploy: "if_needed",
- *   feeMode: "sponsored",
+ *   feeMode: { type: "paymaster" },
  *   onProgress: (e) => console.log(e.step)
  * });
  * ```
@@ -164,11 +170,6 @@ export interface EnsureReadyOptions {
   deploy?: DeployMode;
   /** How to pay for deployment if needed (default: wallet's default) */
   feeMode?: FeeMode;
-  /**
-   * ERC-20 token contract address used to pay deployment gas fees.
-   * @see {@link TransactionFeeOptions.gasToken}
-   */
-  gasToken?: Address;
   /** Callback for progress updates */
   onProgress?: (event: ProgressEvent) => void;
 }
@@ -181,20 +182,6 @@ interface TransactionFeeOptions {
   feeMode?: FeeMode;
   /** Optional time bounds for paymaster transactions */
   timeBounds?: PaymasterTimeBounds;
-  /**
-   * ERC-20 token contract address used to pay gas fees via the paymaster.
-   *
-   * When set, the transaction uses the paymaster with
-   * `{ mode: 'default', gasToken }` instead of full sponsorship.
-   * The user pays gas in the specified token (e.g. the USDC or STRK contract address).
-   *
-   * Requires a paymaster to be configured in the SDK.
-   *
-   * @remarks
-   * - Omit `feeMode` when using `gasToken` (recommended happy path).
-   * - Combining `feeMode: "user_pays"` with `gasToken` throws an error.
-   */
-  gasToken?: Address;
 }
 
 // ─── Deploy ──────────────────────────────────────────────────────────────────
@@ -219,17 +206,14 @@ export interface PreflightOptions {
   /**
    * Fee mode used for preflight assumptions.
    *
-   * When `"sponsored"` and the account is undeployed, preflight returns `{ ok: true }`
+   * When using a paymaster mode (`{ type: "paymaster" }` — with or without
+   * `gasToken`) and the account is undeployed, preflight returns `{ ok: true }`
    * because the paymaster path can deploy + execute atomically.
+   *
+   * The `gasToken` field only affects which token is used for fee payment;
+   * it does not change the preflight deployment decision.
    */
   feeMode?: FeeMode;
-  /**
-   * ERC-20 token contract address used to pay gas fees via the paymaster.
-   *
-   * When set and the account is undeployed, preflight returns `{ ok: true }`
-   * because the paymaster path handles deployment atomically.
-   */
-  gasToken?: Address;
 }
 
 /** Preflight succeeded — operation can proceed */
--- a/src/types/wallet.ts
+++ b/src/types/wallet.ts
@@ -164,49 +164,48 @@ export interface EnsureReadyOptions {
   deploy?: DeployMode;
   /** How to pay for deployment if needed (default: wallet's default) */
   feeMode?: FeeMode;
+  /**
+   * ERC-20 token contract address used to pay deployment gas fees.
+   * @see {@link TransactionFeeOptions.gasToken}
+   */
+  gasToken?: Address;
   /** Callback for progress updates */
   onProgress?: (event: ProgressEvent) => void;
+}
+
+// ─── Transaction Fee Options ─────────────────────────────────────────────────
+
+/** Common fee options shared by deploy and execute operations. */
+interface TransactionFeeOptions {
+  /** How fees are paid (default: "user_pays") */
+  feeMode?: FeeMode;
+  /** Optional time bounds for paymaster transactions */
+  timeBounds?: PaymasterTimeBounds;
+  /**
+   * ERC-20 token contract address used to pay gas fees via the paymaster.
+   *
+   * When set, the transaction uses the paymaster with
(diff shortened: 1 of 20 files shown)


ahmetenesdur:
Hi! Just a gentle ping on this PR in case it got buried.

@0xLucqs @micbakos @0xsisyfos would appreciate a quick review when you have time. Thanks!

0xLucqs (commented):
Sorry for the delay on this review! The feature design is solid — threading `gasToken` through the paymaster path via `sponsoredDetails` is the right approach. Here are the issues I found, ordered by severity.

---

## Bug

### CartridgeWallet.execute() silently ignores explicit `feeMode: "user_pays"` when `gasToken` is set

`Wallet.execute()` correctly guards against `{ feeMode: "user_pays", gasToken: "0x..." }` via `throwGasTokenConflict`, but `CartridgeWallet.execute()` has no such check. Because the routing condition is `feeMode === "sponsored" || gasToken`, a truthy `gasToken` overrides the explicit `"user_pays"` and silently routes to the paymaster:

```typescript
// CartridgeWallet.execute() — current PR code
const feeMode = options.feeMode ?? this.defaultFeeMode; // "user_pays"
const gasToken = options.gasToken;                       // "0x..."
if (feeMode === "sponsored" || gasToken) { ... }         // true — silent override
```

**Fix** — mirror Wallet's validation pattern:
```typescript
const requestedFeeMode = options.feeMode;
const feeMode = requestedFeeMode ?? this.defaultFeeMode;
const gasToken = options.gasToken;

if (requestedFeeMode === "user_pays" && gasToken) {
  throw new Error(
    "Cannot combine feeMode 'user_pays' with gasToken. " +
      "Use feeMode 'sponsored' or omit feeMode when paying with an ERC-20 token."
  );
}
```

Consider extracting this into a shared utility in `utils.ts` so both wallet implementations stay in sync.

---

## Tests

The PR adds zero test coverage. At minimum these scenarios should be covered before merge:

| Scenario | Where |
|---|---|
| `sponsoredDetails()` with gasToken returns `{ mode: "default", gasToken }` | `wallet-utils.test.ts` |
| `sponsoredDetails()` without gasToken returns `{ mode: "sponsored" }` | `wallet-utils.test.ts` |
| `Wallet.execute({ gasToken, feeMode: "user_pays" })` throws | `wallet.test.ts` |
| `Wallet.deploy({ gasToken, feeMode: "user_pays" })` throws | `wallet.test.ts` |
| `gasToken` without explicit feeMode routes to paymaster | `wallet.test.ts` |
| `CartridgeWallet.deploy({ gasToken })` throws | `cartridge.test.ts` |
| `CartridgeWallet.execute({ gasToken })` routes to paymaster | `cartridge.test.ts` |

Note: `sponsoredDetails()` currently has zero test coverage (even before this PR), so adding baseline tests would be valuable.

---

## Suggestions (non-blocking)

### 1. Extract shared base type for DeployOptions / ExecuteOptions

Both interfaces are now structurally identical. A shared base eliminates the JSDoc duplication and prevents future field drift:

```typescript
interface TransactionFeeOptions {
  feeMode?: FeeMode;
  timeBounds?: PaymasterTimeBounds;
  gasToken?: Address;
}
export interface DeployOptions extends TransactionFeeOptions {}
export interface ExecuteOptions extends TransactionFeeOptions {}
```

### 2. Rename `sponsoredDetails` → `paymasterDetails`

The function now returns `{ mode: "default" }` when `gasToken` is set, which is not "sponsored". `paymasterDetails` or `buildPaymasterDetails` would be accurate for both code paths.

### 3. Consider an options object for `sponsoredDetails`

The function now takes 3 positional optionals `(timeBounds?, deploymentData?, gasToken?)`, which forces callers to pass `undefined` as a placeholder (e.g. `sponsoredDetails(timeBounds, undefined, gasToken)` in CartridgeWallet). An options object would be cleaner:

```typescript
function paymasterDetails(options: {
  timeBounds?: PaymasterTimeBounds;
  deploymentData?: PAYMASTER_API.ACCOUNT_DEPLOYMENT_DATA;
  gasToken?: Address;
} = {})
```

### 4. Clarify gasToken JSDoc

The example `(e.g. USDC, STRK)` could be misread as token symbols. Since the type is `Address`, something like `(e.g. the USDC or STRK contract address)` would be clearer.

Also worth documenting the `feeMode` interaction — specifically that omitting `feeMode` with `gasToken` is the intended happy path, but combining `feeMode: "user_pays"` with `gasToken` throws.

### 5. `ensureReady` doesn't support gasToken

`ensureWalletReady()` calls `wallet.deploy(feeMode ? { feeMode } : undefined)` — no way to pass `gasToken`. If gasToken-paid deployment via `ensureReady` is a valid use case, `EnsureReadyOptions` should also get the `gasToken` field.

---

Overall this is a clean, well-scoped feature. The CartridgeWallet validation gap and test coverage are the two items I'd want addressed before merge. The rest are optional improvements. Nice work!

ahmetenesdur:
@0xLucqs Thanks for the review! All feedback has been addressed:

- Added the missing conflict guard to `CartridgeWallet.execute()`
- Renamed `sponsoredDetails` → `paymasterDetails` with options object pattern
- Extracted conflict guard to shared `utils.ts`
- Introduced `TransactionFeeOptions` base type, clarified JSDoc
- Added `gasToken` to `EnsureReadyOptions` and `PreflightOptions`
- Full test coverage for all changes

All 550 tests passing.

micbakos (commented):
@ahmetenesdur My only remark regards to the structure of the input of `FeeMode`. Gas token can only be used with paymaster. So it is possible a user to provide `feeMode: "user_pays", gasToken: "0x..."`. In this case gasToken will be ignored at runtime but again it implies that you can pay any tx with any gas token. So instead I would argue that we could improve on the design of the type like so:

```
/**
 * How transaction fees are paid.
 * - `{ type: "paymaster" }`: Paymaster covers gas with its default token
 * - `{ type: "paymaster", gasToken }`: Paymaster uses the specified gas token
 * - `"user_pays"`: User's account pays gas in ETH/STRK
 * - `"sponsored"`: Deprecated alias for paymaster-covered gas
 */
/**
 * @deprecated Use `{ type: "paymaster" }` instead.
 */
type DeprecatedSponsoredFeeMode = "sponsored";

export type FeeMode =
  | DeprecatedSponsoredFeeMode
  | {
      type: "paymaster";
      gasToken?: Token | Address | undefined;
    }
  | "user_pays";
```
Now when "paymaster" is defined you can optionally attach a gasToken token or address or ignore it completely. Keep in mind that I kept "sponsored" as a possible type **but deprecated** so as not to break compatibility with older versions.

ahmetenesdur:
@micbakos Good point — making the invalid state unrepresentable at the type level is cleaner than a runtime guard.

A couple of things to consider though:

1. **starknet.js alignment**: The underlying `FEE_MODE` type from [`starknet-io/types-js`](https://github.com/starknet-io/types-js/blob/main/src/snip-29/components.ts#L90-L97) already uses `{ mode: 'sponsored' } | { mode: 'default', gas_token }`. If we restructure our FeeMode, we'd want to align with that shape rather than introducing a new `type: "paymaster"` discriminant.

2. **Scope**: This would touch ~13 files and all routing logic — better suited as a dedicated follow-up PR.

The current runtime guard (`assertNoGasTokenConflict`) catches this at all entry points. Happy to open an issue for the type-level redesign as a follow-up.

micbakos:
> @micbakos Good point — making the invalid state unrepresentable at the type level is cleaner than a runtime guard.
> 
> A couple of things to consider though:
> 
> 1. **starknet.js alignment**: The underlying `FEE_MODE` type from [`starknet-io/types-js`](https://github.com/starknet-io/types-js/blob/main/src/snip-29/components.ts#L90-L97) already uses `{ mode: 'sponsored' } | { mode: 'default', gas_token }`. If we restructure our FeeMode, we'd want to align with that shape rather than introducing a new `type: "paymaster"` discriminant.
> 2. **Scope**: This would touch ~13 files and all routing logic — better suited as a dedicated follow-up PR.
> 
> The current runtime guard (`assertNoGasTokenConflict`) catches this at all entry points. Happy to open an issue for the type-level redesign as a follow-up.

I am not fully convinced that this is the correct approach. This is a higher level library, and not a client built to interface with starknet rpc directly. There is no `"user_pays"` either in starknet.js. The code you are referring to regards to paymaster fee mode. In the grand scheme of things a mode called "default" would be counter intuitive to people using the `execute` function. The reason we suggest using 'paymaster' instead of just 'sponsored' is that sponsored means that someone pays for you. So applying a `gasToken` parameter next to 'sponsored' makes no sense (why do I need a a gasToken if someone else pays for my txs?). So a more "generic" `'paymaster'` with an optional `gas_token` type would fit better in this scenario.

ahmetenesdur:
@micbakos You're right — `"paymaster"` with an optional `gasToken` makes more sense semantically at the SDK level. I'll update the PR with this refactor.

ahmetenesdur:
@micbakos Refactor pushed in `8dc9a60`.

FeeMode is now `"user_pays" | { type: "paymaster"; gasToken?: Address } | "sponsored"` — the deprecated `"sponsored"` alias is kept for backward compat.

Removed `assertNoGasTokenConflict()` since the type system now prevents invalid combos. Added `normalizeFeeMode()` and `isPaymasterMode()` for clean internal routing. 20 files touched, 740 tests passing.

src/types/wallet.ts
```diff
@@ -65,10 +65,16 @@ export interface AccountConfig {
...
+ * - `"user_pays"` — User's account pays gas in ETH/STRK
+ * - `{ type: "paymaster" }` — Paymaster covers gas (sponsored)
+ * - `{ type: "paymaster", gasToken: "0x..." }` — Pay gas via ERC-20 through paymaster
+ * - `"sponsored"` — *(deprecated)* Alias for `{ type: "paymaster" }`
  */
-export type FeeMode = "sponsored" | "user_pays";
+export type FeeMode =
+  | "user_pays"
+  | { type: "paymaster"; gasToken?: Address }
+  | "sponsored";
```
micbakos: Please adhere to the suggestion proposed in https://github.com/keep-starknet-strange/starkzap/pull/119#pullrequestreview-4120117705

```suggestion
  | DeprecatedSponsoredFeeMode
  
/**
 * @deprecated Use `{ type: "paymaster" }` instead.
 */
type DeprecatedSponsoredFeeMode = "sponsored";
```

`@deprecated` annotation is needed to warn users not to use it anymore

0xLucqs: also i don't think cartridge should accept `gasToken`

ahmetenesdur:
@micbakos @0xLucqs Both addressed in `c4ee643`:

- `"sponsored"` → `DeprecatedSponsoredFeeMode` type alias with `@deprecated` JSDoc
- Native Cartridge now rejects `gasToken` at validation time instead of silently ignoring it

All 740 tests pass, lint/formatting clean.

0xLucqs (approved):

micbakos (approved):

PR #129, opened 2026-05-20
Layerswap integration

babkenmes:
## Summary

Integrates [Layerswap](https://www.layerswap.io/) as a new bridge provider, expanding StarkZap's cross-chain coverage alongside the existing Canonical, CCTP, OFT, and Hyperlane providers. Layerswap brings:

- **Ethereum ↔ Starknet** deposits and withdrawals
- **Solana ↔ Starknet** deposits and withdrawals
- Lifecycle monitoring (status polling) consistent with other providers
- Quote/fee estimation via the live Layerswap API

No new runtime dependencies — the Layerswap HTTP client is inlined.

## What's added

- **`Protocol.LAYERSWAP`** added to the `Protocol` enum and to both `EthereumBridgeProtocol` and `SolanaBridgeProtocol` unions.
- **`BridgingConfig`** gains `layerSwapApiKey` and optional `layerSwapBaseUrl`. The API key is required when bridging through Layerswap; `BridgeOperator` throws a clear error if the protocol is routed without one.
- **`LayerswapBridge`** (`src/bridge/ethereum/layerswap/`) — extends `EthereumBridge`. The deposit flow creates a swap on the Layerswap API, executes the returned EVM deposit action via the connected signer, and pings `speedUpDeposit` for faster detection. Withdrawals trigger Starknet→Ethereum transfers through prebuilt `call_data`.
- **`SolanaLayerswapBridge`** (`src/bridge/solana/`) — implements `BridgeInterface<SolanaAddress>` and sends Layerswap's prebuilt Solana transaction through the Reown-compatible provider.
- **`LayerswapMonitor`** (`src/bridge/monitor/layerswap/`) — polls Layerswap swap status and maps it to the SDK's shared deposit/withdrawal state model.
- **Fee estimation types**:
  - `LayerswapDepositFeeEstimation` (Ethereum) extends `EthereumDepositFeeEstimation` with `serviceFee`, `receiveAmount`, `avgCompletionTime`.
  - `SolanaLayerswapDepositFeeEstimation` is a flat Solana-side shape; the SDK-level `BridgeDepositFeeEstimation` union enumerates Ethereum, Hyperlane, and Layerswap-Solana variants directly.
- **Web example** (`examples/web/bridge/`) updated with a unified fee-estimate formatter for the new variants and stubbed Layerswap tokens (see below).

## Token discovery and stubbed tokens ⚠️

Layerswap tokens are designed to flow through the existing `BridgeTokenRepository` pipeline once the StarkGate token API ([starkgate.starknet.io/tokens/api/tokens](https://starkgate.starknet.io/tokens/api/tokens)) starts returning entries with `protocol: "layerswap"`. The repository parser already accepts the new value with no further SDK changes required.

**Because StarkGate doesn't yet serve `protocol: "layerswap"` records, the web example temporarily injects hardcoded stub tokens** so the integration is testable end-to-end:

- `examples/web/bridge/index.ts` → `getEthereumLayerswapTestTokens()` and `getSolanaLayerswapTestTokens()` add ETH/USDC/etc. tokens with `Protocol.LAYERSWAP` when the API returns none.
- These stubs are scoped to the web example only — the SDK itself does not hardcode any Layerswap tokens. Production consumers will get Layerswap tokens automatically once the StarkGate API ships them.
- **Follow-up:** once StarkGate adds Layerswap tokens, delete the two stub functions and the call sites in `fetchTokens()`. No SDK changes will be needed.

## Environment variables (web example)

The web example (`examples/web`) reads one new Vite env var at startup:

| Variable | Purpose | Required for |
|---|---|---|
| `VITE_LAYERSWAP_API_KEY` | API key passed to the Layerswap REST API | Any Layerswap bridging in the example (deposit, withdrawal, quotes) |

It maps to a `BridgingConfig` field:

```ts
new StarkZap({
  // ...
  bridging: {
    layerswapApiKey: VITE_LAYERSWAP_API_KEY, // required when Protocol.LAYERSWAP is routed
  },
});

### CODE REVIEW

File: src/bridge/ethereum/layerswap/LayerswapBridge.ts [hunk 1]
```diff
@@ -0,0 +1,414 @@
...
+ * 1. Creates a swap on Layerswap API
+ * 2. Retrieves deposit actions (EVM transactions to execute)
+ * 3. Executes the deposit on Ethereum via the connected signer
+ * 4. Notifies Layerswap for faster detection
+ *
+ * Routed by {@link BridgeOperator} when `token.protocol === Protocol.LAYERSWAP`.
+ */
+export class LayerswapBridge extends EthereumBridge {
+  private readonly api: LayerswapApi;
+  private readonly sourceNetwork: string;
```
micbakos: I would rename this to `evmNetwork`

src/bridge/ethereum/layerswap/LayerswapBridge.ts
```diff
@@ -0,0 +1,414 @@
...
+  private readonly destNetwork: string;
```
micbakos: And this `starknetNetwork`. Both are not always src or dest but can be reversed.

src/bridge/ethereum/EtherToken.ts
```diff
@@ -38,11 +38,18 @@ export type EthereumTokenInterface = {
   isNativeEth(): boolean;
 };
 
+const NATIVE_ETH_ADDRESS = "0x0000000000000000000000000000000000000000";
```
micbakos: 👍🏼

src/bridge/monitor/layerswap/LayerswapMonitor.ts
```diff
@@ -0,0 +1,197 @@
...
+        return BridgeTransferStatus.SUBMITTED_ON_L1;
+      case "ls_transfer_pending":
+        return BridgeTransferStatus.CONFIRMED_ON_L1;
+      case "completed":
+        return BridgeTransferStatus.CONFIRMED_ON_STARKNET;
+      case "failed":
+      case "expired":
+      case "cancelled":
+      case "pending_refund":
+      case "refunded":
```
micbakos, line 174: Minor: Here a default case that could be resulting in Error (alongside an appropriate log) would suffice. Just to make sure the monitoring functions do not have weird results if LS status changes.

src/bridge/monitor/layerswap/LayerswapMonitor.ts
```diff
@@ -0,0 +1,197 @@
...
+      case "ls_transfer_pending":
+        return BridgeTransferStatus.CONFIRMED_ON_STARKNET;
+      case "completed":
+        return BridgeTransferStatus.COMPLETED_ON_L1;
+      case "failed":
+      case "expired":
+      case "cancelled":
+      case "pending_refund":
+      case "refunded":
+        return BridgeTransferStatus.ERROR;
```
micbakos, line 201: Same as above

src/bridge/solana/SolanaLayerswapBridge.ts
```diff
@@ -0,0 +1,377 @@
...
+    private readonly logger: StarkZapLogger,
+    apiConfig?: Omit<LayerswapApiConfig, "apiKey">
+  ) {
+    this.api = new LayerswapApi({ apiKey, ...apiConfig });
+    this.starknetToken = new Erc20(
+      bridgeToken.intoStarknetToken(),
+      starknetWallet.getProvider()
+    );
+    const mainnet = starknetWallet.getChainId().isMainnet();
+    this.sourceNetwork = mainnet ? "SOLANA_MAINNET" : "SOLANA_DEVNET";
```
micbakos: Is the solana test network available on LS be SOLANA_DEVNET only? BridgeOperator currently connects to `testnet` as seen [here](https://github.com/keep-starknet-strange/starkzap/blob/256e1fb3735567bf8fb96f047c6217ef2781f74a/src/bridge/operator/BridgeOperator.ts#L534) (due to how Hyperlane is integrated). If LS does not support testnet then `BridgeOperator.getSolanaConnection` needs to be updated with a cluster param. The cluster param also needs to be checked against current starknet wallet chain id.

src/types/config.ts
```diff
@@ -178,6 +178,20 @@ export interface BridgingConfig {
...
+   * Required only when bridging tokens via Layerswap.
+   */
+  layerswapApiKey?: string;
+
+  /**
+   * Custom Layerswap API base URL.
+   *
+   * Defaults to `https://api.layerswap.io` when omitted.
+   */
+  layerswapBaseUrl?: string;
```
micbakos, line 195: This url needs to be validated at runtime. Other urls are run through `assertSafeHttpUrl`.

micbakos (commented):
Thank you for the PR. I have some minor takes on it. Besides that I would like to suggest a solution to the current stubbing problem with available bridge tokens:

**Point 1**
`BridgeTokenRepository` discovers Layerswap tokens when an API key is configured

The repository takes the Layerswap API key as an optional dependency. When it's present, `getTokens` also fetches Layerswap routes and merges them into the result. When it's absent, behavior is identical to today so zero impact for non-Layerswap consumers.  `getTokens(query)` keeps its exact signature, so this is fully caller-transparent.

The route parameters aren't exposed to the caller they're derived from what BridgeTokenQuery already carries. Because Layerswap in this SDK always bridges with Starknet on one side and the only external chains are Ethereum and Solana, the discovery calls are a deterministic function of (env, chain):

| `chain`     | `env`   | external network   | Starknet counterpart | `networkTypes`        |
  | ----------- | ------- | ------------------ | -------------------- | --------------------- |
  | ETHEREUM    | mainnet | `ETHEREUM_MAINNET` | `STARKNET_MAINNET`   | `["evm"]`             |
  | ETHEREUM    | testnet | `ETHEREUM_SEPOLIA` | `STARKNET_SEPOLIA`   | `["evm"]`             |
  | SOLANA      | mainnet | `SOLANA_MAINNET`   | `STARKNET_MAINNET`   | `["solana"]`          |
  | SOLANA      | testnet | `SOLANA_DEVNET` ⚠️  | `STARKNET_SEPOLIA`   | `["solana"]`          |
  | _(undefined)_ | …     | both               | per env              | `["evm","solana"]`    |

 So the calls become hardcoded-from-the-caller's-view:
  
```
// external → Starknet: external token contract, symbol, decimals
api.getSources({ destinationNetwork: starknetNetwork, networkTypes });
// Starknet-side tokens for the same route → match by symbol to fill starknetAddress
api.getDestinations({ sourceNetwork: externalNetwork });
```
**Point 2**
Split the `BridgeToken` hierarchy so bridge contract addresses are canonical only.
Now stubbed Layerswap tokens are modeled by passing zero addresses (`l1Bridge: 0x0…`, `starknetBridge: 0x0`) into required fields. 

Lift those two fields off the shared base:
```
BridgeToken (base)         id, name, symbol, decimals, address, starknetAddress, protocol, chain
    ├─ CanonicalBridgeToken   + bridgeAddress + starknetBridge   (canonical / cctp / oft / hyperlane / lords)
    └─ LayerswapBridgeToken   (no bridge-contract addresses — deposit address is per-swap from the API)
```
Lastly `LayerswapBridge extends EthereumBridge`, whose constructor eagerly does `new Contract(bridgeToken.bridgeAddress)`, so every Layerswap bridge currently instantiates a dead ethers Contract at the zero address. Moving that construction down into `CanonicalEthereumBridge` removes the fragility.

Let me know what you think.

src/bridge/solana/SolanaLayerswapBridge.ts
babkenmes: Unfortunately Layerswap only supports Solana devnet (SOLANA_DEVNET).

babkenmes:
Thanks a lot for the thorough review @micbakos really appreciate the suggestions, especially the discovery design and the token-hierarchy split. I've addressed the points:
- Token discovery (Point 1): Implemented as you described. BridgeTokenRepository takes the Layerswap API key as an optional dependency and discovers routes via getSources/getDestinations when it's set, derived from (env, chain) through a single resolveLayerswapRoute() helper. 
- BridgeToken hierarchy (Point 2): Done and generalized. Bridge-contract addresses are lifted off the base BridgeToken/EthereumBridgeToken/SolanaBridgeToken into ContractRouted* subclasses, which carry bridgeAddress/starknetBridge. The eager new Contract(...) now lives only in ContractRoutedEthereumBridge (extended by CanonicalEthereumBridge for canonical/lords and by OftBridge). The two protocols that resolve their contracts elsewhere Layerswap (per-swap deposit address from the API) and CCTP (TokenMessenger/MessageTransmitter from chain constants) extend the base EthereumBridge directly and parse to the plain base token classes, so neither instantiates a dead Contract at a zero address. The distinction is encoded in isContractRouted() and enforced by the type system rather than asserted at runtime.
- Renames: sourceNetwork/destNetwork → evmNetwork/starknetNetwork (and solanaNetwork/starknetNetwork on the Solana side).
- Monitor status switches: Both now have a default branch that logs a warning and returns ERROR, so an LS status change can't produce a silently-wrong result.
- layerswapBaseUrl: Validated at runtime via assertSafeHttpUrl in the LayerswapApi constructor (every construction path goes through it).

micbakos:
Hey @babkenmes just to make sure we don't just push a version that doesn't work to our users, I noticed when testing the intergation using `examples/web` (after setting an api key) that both `sources` and `destinations` endpoints return 404 thus making the discovery of tokens impossible:

```
https://api.layerswap.io/api/v2/sources?destination_network=STARKNET_SEPOLIA&network_types=evm
// and
https://api.layerswap.io/api/v2/destinations?source_network=ETHEREUM_SEPOLIA
```
both respond with 404.

babkenmes:
> Hey @babkenmes just to make sure we don't just push a version that doesn't work to our users, I noticed when testing the intergation using `examples/web` (after setting an api key) that both `sources` and `destinations` endpoints return 404 thus making the discovery of tokens impossible:
> 
> ```
> https://api.layerswap.io/api/v2/sources?destination_network=STARKNET_SEPOLIA&network_types=evm
> // and
> https://api.layerswap.io/api/v2/destinations?source_network=ETHEREUM_SEPOLIA
> ```
> 
> both respond with 404.

Hey @micbakos what api key are you using? Layerswap provides separate api keys for mainnet and testnet routes
An api key for testnet `Dz1jVir9WUD0gBWoGbOmS1oe5K4985SGptaZXjF4z9VVrvO5nC9q55h8TE/3CIESRxWdYVpPnz/H2BogL2eG+A`

micbakos:
> Dz1jVir9WUD0gBWoGbOmS1oe5K4985SGptaZXjF4z9VVrvO5nC9q55h8TE/3CIESRxWdYVpPnz/H2BogL2eG+A

Ahh sorry for that, testing again

micbakos:
@babkenmes Still getting 404. Am I missing somthing? Integrated `VITE_LAYERSWAP_API_KEY` using your key. Does testnet need a different base url?

babkenmes:
@micbakos fixed the api key is now passed to the route discovery, but this rises an issue, users will not be able to use Layerswap without api key in testnet as keyless only works in mainnet. 
We can make the api key required for Layerswap and always get the routes from Layerswap API. What do you think of this?

micbakos:
> ers will not be able to use Layerswap without api key in testnet as keyless only works in mainnet.

I think making the API key required makes sense as it follows the same pattern as the rest of the integrations.

micbakos:
Took the liberty to fix what I meant earlier so as to save some time in back and forth:

###  Layerswap token discovery is now optional
  This is what I meant in the begining.  Previously, the SDK required a Layerswap API key and would fail if one wasn't configured. Now the key is optional:
  - When a key is provided, the SDK discovers Layerswap-bridgeable tokens as before.
  - When no key is provided, the SDK simply skips Layerswap discovery and continues to return the standard StarkGate bridge tokens — no error is raised.
  
  This makes the SDK usable for bridging scenarios that don't involve Layerswap, without forcing every integrator to obtain a Layerswap key.

### Clearer errors for Solana network/protocol mismatches on testnet

  On Starknet's test environment (Sepolia), the two supported Solana bridge protocols operate on different Solana networks:
  - Layerswap bridges via Solana Devnet.
  - Hyperlane bridges via Solana Testnet.
  
  The SDK now validates this automatically and raises a clear, descriptive error if the connected Solana wallet is on the wrong network for the chosen protocol. It also connects to the correct Solana cluster for each protocol behind the scenes, removing a previously silent source of failures.

### Support for Solana Devnet wallets
  
  The SDK now recognizes and accepts connections from Solana Devnet wallets (in addition to Mainnet and Testnet), with validation ensuring Devnet is only used together with Starknet Sepolia. This is what enables Layerswap bridging on the test environment.

  Supporting changes

  - The bundled web example was updated to allow connecting Solana Devnet wallets and to select the appropriate environment-specific Layerswap API key (Layerswap issues separate keys for mainnet
  and testnet).
  - Test coverage was updated to reflect the new optional-key behavior.

 ----

@babkenmes Something that is still missing. When connecting (to the web example) to a solana devnet wallet the token discovery for solana **fails with 404**. The rest of the scenarios work nicely.

Specifically 
```
https://api.layerswap.io/api/v2/destinations?source_network=SOLANA_DEVNET => returns 404
https://api.layerswap.io/api/v2/sources?destination_network=STARKNET_SEPOLIA&network_types=solana => returns 404
```
Beware that you will need to setup `VITE_LAYERSWAP_API_KEY_MAINNET` and `VITE_LAYERSWAP_API_KEY_TESTNET` on the web example's `.env` so it can work on both starknet mainnet and sepolia connected wallets.

babkenmes:
@micbakos Thanks a lot for taking care of these changes directly really appreciate the help.
The Solana Devnet route is now available as well.

micbakos (approved):
The integration seems to be working @babkenmes

PR #130, opened 2026-06-12
Reject overprecise Amount scalars

samsamtrum:
Amount multiply and divide silently truncated scalar fractions beyond 18 decimals, which can turn a precise input into a different value.

This rejects scalars above the supported precision and keeps scientific notation normalization consistent with Amount parsing.


micbakos:
Hey @samsamtrum thanks for the fix. Applied a fix on a failing test, added more tests related to precision, removed a specific close-to-zero case on division since it is not needed after your changes and improve errors.

micbakos (approved):

PR #136, opened 2026-08-03
fix: make optional peer deps truly optional (lazy-import tongo-sdk/ethers in root-reachable modules)

l-henri:
## Problem

starkzap declares its heavy deps as **optional** peerDependencies (`@fatsolutions/tongo-sdk`, `ethers`, `@solana/web3.js`, `@hyperlane-xyz/*`, `@avnu/avnu-sdk`, `@cartridge/controller`) and already has lazy runtime loaders for most of them (`loadEthers`, `loadSolanaWeb3`, `loadHyperlane`, `loadAvnuSdk`) — clearly the intended design. But in practice a consumer that uses none of those features still **cannot build** without installing the peers, for two reasons:

**1. Static peer imports survive in the emitted JS.** In the current `main` build:

- `dist/src/confidential/tongo.js` → `import { Account as TongoAccount, pubKeyBase58ToAffine } from "@fatsolutions/tongo-sdk"`, statically reachable from the root: `index.js → confidential/index.js → confidential/tongo.js`. Every bundler must resolve tongo-sdk just to import `StarkZap`.
- Six bridge modules import `ethers` values at module scope: `bridge/ethereum/{EthereumBridge,EtherToken,ContractRoutedEthereumBridge}.js`, `bridge/ethereum/cctp/CCTPBridge.js`, `bridge/ethereum/oft/LayerZeroApi.js`, `bridge/monitor/canonical/utils.js` — plus `bridge/ethereum/canonical/CanonicalEthereumBridge.js`, where the inline-type form `import { type X } from "ethers"` still emits `import {} from "ethers"` under `verbatimModuleSyntax`. These are reached through `BridgeOperator`'s dynamic `import()`s — which bundlers **also resolve at build time** — so they fail consumer builds even though they're off the static root graph.

**2. The lazy loaders themselves hard-fail bundlers.** The `.then()/.catch()` chain shape (`import("ethers").then(...).catch(...)`) is treated by webpack as an unresolvable request → **build ERROR**. Only a `try { await import("x") } catch` shape (the one `src/wallet/cartridge.ts` already uses) is downgraded to a build warning + run-time error. esbuild behaves the same, and additionally needs each `import()` to be directly awaited in the `try` (the `Promise.all([import(...), ...])` shape in `hyperlaneRuntime` still hard-fails).

### Real-world impact

A Next.js 15 app importing only `StarkZap` + `PrivySigner` fails to build with 5 "Module not found" errors. `webpack.IgnorePlugin` is **not** a workaround: ignored modules throw at load time, which kills `import("starkzap")` at runtime — consumers are forced to alias each peer to a hand-written Proxy stub. Reproduced here with vanilla webpack 5 against the packed 3.0.0-main tarball in a project with no optional peers installed: **7 build errors** (3× hyperlane, tongo-sdk, ethers, @solana/web3.js, @avnu/avnu-sdk).

## Approach

Follow the codebase's own lazy-runtime pattern everywhere; no subsystem was moved to a subpath export.

- **`src/confidential/tongo.ts`** — new `src/confidential/tongoRuntime.ts` (`loadTongo`/`requireTongo`, mirroring `ethersRuntime`). `TongoConfidential`'s constructor now kicks off the lazy load and builds the underlying `Account` when it resolves; all async methods await it. New public `ready(): Promise<void>`.
- **`src/connect/ethersRuntime.ts`** — added a sync `requireEthers(feature)` accessor for code paths that can't await (constructors, static helpers). It's safe wherever the caller already holds an ethers object (`Provider`, `Signer`, receipt — which can't exist unless ethers loaded) and in bridge classes, which are only constructed after `BridgeOperator` awaits `loadEthers`.
- **Bridge modules** — all `ethers` imports are now `import type`; value uses go through `await loadEthers(...)` in async methods, `requireEthers(...)` in sync ones (each site has a comment justifying why sync access is safe there). `CCTPBridge`'s module-eval statics (`new Interface(...)`, checksummed messenger addresses) became lazy cached static getters.
- **All five runtime loaders** (`ethers`, `@solana/web3.js`, hyperlane, avnu, tongo) — restructured from `.then()/.catch()` chains to `try { await import(...) } catch`, with hyperlane's three imports awaited sequentially, so webpack/esbuild downgrade a missing peer to the intended run-time install-hint error instead of a build failure. Error messages are unchanged.

### Regression guards (new tests)

- `tests/optional-peer-imports.test.ts` — walks the emitted-JS module graph from `src/index.ts`/`src/cartridge.ts` (static imports/re-exports **and** internal dynamic `import()`s, modeling `verbatimModuleSyntax` elision rules) and asserts no optional peer is imported statically. It fails on current `main` with exactly the 8 offending files.
- `tests/tongo-optional-dep.test.ts` — mirrors the existing `avnu-optional-dep.test.ts`: with tongo-sdk missing, construction doesn't throw and `ready()`/async methods reject with the install hint.

## API-visible consequences

- `TongoConfidential` gains `ready(): Promise<void>`.
- `TongoConfidential`'s **sync** accessors (`address`, `recipientId`, `recipientFromAddress`, `getTongoAccount`) now require the lazy module load to have settled — i.e. `await ready()` (or any async method) first — and otherwise throw an actionable error. This is the one behavior change: previously `new TongoConfidential(...).address` worked synchronously. It could not be preserved without a static import (top-level await was rejected for bundler-compat reasons). All async methods behave exactly as before.
- Everything else is internal; no exports were added/removed from `src/index.ts`.

## Verified

- `npm run typecheck`, `npm run typecheck:all`, `npm run lint`, `npm run prettier:check`, `npm run build`, `npm run build:all`: pass.
- `npm test`: 762 passed / 12 skipped; the only 2 failures (`tests/amount.test.ts` thousand-separator cases) also fail on a clean `main` checkout in this environment (locale/ICU-dependent), unrelated to this change.
- Integration tests were **not** run (they need a local devnet/fork setup).
- **End-to-end no-peers repro**: `npm pack` the built package → install the tarball into a fresh project with no optional peers →
  - `node --input-type=module -e "await import('starkzap')"` evaluates; `new TongoConfidential(...)` then `await ready()` rejects with the `npm i @fatsolutions/tongo-sdk` hint;
  - **webpack 5** production build of an app importing `StarkZap` + `PrivySigner`: **0 errors** (down from 7), only the expected "Module not found" *warnings* on the lazy loaders; the emitted bundle evaluates and both exports are usable;
  - **esbuild** `--bundle` of the same app: exits 0; the bundle runs.
  - Before the loader restructuring, the same repro produced 7 webpack ERRORs, confirming both halves of the fix are needed.

## Notes for reviewers

- The `requireEthers` sync-accessor sites are the part worth scrutinizing: they assume ethers is already cached when reached (arguments given in each site's comment). If you'd rather pay the churn, any of them can be converted to async + `await loadEthers`.
- The sequential (vs `Promise.all`) hyperlane imports trade a little first-load parallelism for esbuild compatibility.
- Docs under `docs/api/**` are generated and not tracked in git, so they were not regenerated here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/bridge/ethereum/ContractRoutedEthereumBridge.ts
+++ b/src/bridge/ethereum/ContractRoutedEthereumBridge.ts
@@ -14,7 +14,8 @@ import type {
   EthereumTransactionDetails,
   EthereumWalletConfig,
 } from "@/bridge/ethereum/types";
-import { Contract, type ContractTransaction, type InterfaceAbi } from "ethers";
+import type { Contract, ContractTransaction, InterfaceAbi } from "ethers";
+import { requireEthers } from "@/connect/ethersRuntime";
 import { FeeErrorCause } from "@/types/errors";
 import type { WalletInterface } from "@/wallet";
 import CANONICAL_BRIDGE_ABI from "@/abi/ethereum/canonicalBridge.json";
@@ -45,6 +46,10 @@ export abstract class ContractRoutedEthereumBridge extends EthereumBridge {
     bridgeAbi: InterfaceAbi = CANONICAL_BRIDGE_ABI
   ) {
     super(bridgeToken, config, starknetWallet, logger);
+    // Sync accessor is safe: bridges are only constructed after
+    // `BridgeOperator.createEthereumBridge` has awaited `loadEthers`, and the
+    // `config.signer` passed in is itself an ethers object.
+    const { Contract } = requireEthers("Ethereum bridge construction");
     this.bridge = new Contract(
       bridgeToken.bridgeAddress,
       bridgeAbi,
--- a/src/bridge/ethereum/EtherToken.ts
+++ b/src/bridge/ethereum/EtherToken.ts
@@ -3,20 +3,19 @@ import {
   ExternalChain,
   NATIVE_TOKEN_ADDRESS,
 } from "@/types/bridge/external-chain";
-import {
-  Contract,
-  type ContractTransaction,
-  getAddress,
-  type Provider,
-  type Signer,
-} from "ethers";
+import type { Contract, ContractTransaction, Provider, Signer } from "ethers";
 import ERC20_ABI from "@/abi/ethereum/erc20.json";
 import { type EthereumWalletConfig } from "@/bridge/ethereum/types";
-import { fromEthereumAddress } from "@/connect/ethersRuntime";
+import {
+  fromEthereumAddress,
+  loadEthers,
+  requireEthers,
+} from "@/connect/ethersRuntime";
 
 export async function ethereumAddress(
   contract: Contract
 ): Promise<EthereumAddress> {
+  const { getAddress } = await loadEthers("Ethereum token address");
   const target = contract.target;
   const address =
     typeof target === "string" ? target : await target.getAddress();
@@ -65,6 +64,9 @@ export class ERC20EthereumToken implements EthereumTokenInterface {
   }>;
 
   public static create(address: EthereumAddress, provider: Provider) {
+    // Sync accessor is safe: the caller holds an ethers `Provider`, which
+    // cannot exist unless ethers has already been loaded.
+    const { Contract } = requireEthers("Ethereum ERC20 token");
     const contract = new Contract(address, ERC20_ABI, provider);
     return new ERC20EthereumToken(contract);
   }
--- a/src/bridge/ethereum/EthereumBridge.ts
+++ b/src/bridge/ethereum/EthereumBridge.ts
@@ -18,18 +18,15 @@ import {
   type ApprovalFeeEstimation,
   type EthereumWalletConfig,
 } from "@/bridge/ethereum/types";
-import {
-  type ContractTransaction,
-  type ContractTransactionReceipt,
-  type ContractTransactionResponse,
-  getAddress,
-  isError,
-  toBigInt,
-  type TransactionRequest,
+import type {
+  ContractTransaction,
+  ContractTransactionReceipt,
+  ContractTransactionResponse,
+  TransactionRequest,
 } from "ethers";
 import { FeeErrorCause, TransactionErrorCause } from "@/types/errors";
 import type { WalletInterface } from "@/wallet";
-import { fromEthereumAddress } from "@/connect/ethersRuntime";
+import { fromEthereumAddress, loadEthers } from "@/connect/ethersRuntime";
 import { Erc20 } from "@/erc20";
 import { type StarkZapLogger } from "@/logger";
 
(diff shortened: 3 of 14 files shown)


micbakos:
Hey @l-henri thanks for the PR:

* Part of this PR is fixed in #135, related to tongo's new `await ready()` method. In 135 something like that is not needed since in order to get a new instance you have to go through the `create(...)` factory which returns a promise. So no need for the `ready()` method.
* But in terms of the webpack errors, you are right and they are welcomed to be merged.
* Also inspired by this PR I also fixed the dynamic import of `privacy-sdk` in 135.

So I would suggest, 
1. first waiting for 135 to be merged, 
2. then rebase this branch against updated main and remove the `ready()` functionality since it will not be needed, otherwise we would also need to update the examples based on the updated API.

I think you AI could review this commit and let me know if it agrees.

micbakos:
Rebased this branch onto `main` and dropped the parts that #140 superseded, so it can go in as-is. Force-pushed, so `git fetch && git reset --hard origin/fix/optional-peers-lazy-imports` if you have it checked out locally.

**Kept unchanged** (your commit, `a732791`): `ethers` out of the root graph across the seven bridge modules, `requireEthers` for the sync call sites, `CCTPBridge`'s lazily cached statics, the `try { await import() } catch` restructuring of the avnu / ethers / solana / hyperlane loaders, and `tests/optional-peer-imports.test.ts`.

**Dropped**, because #140 landed in the meantime and solves the same problem differently:

- `src/confidential/tongoRuntime.ts` and `TongoConfidential.ready()` — #140 made `TongoConfidential.create(...)` an async factory, so the module is loaded before an instance exists. The synchronous accessors (`address`, `recipientId`, `recipientFromAddress`, `getTongoAccount`) keep working with no `ready()` step and never observe a half-built object.
- `tests/tongo-optional-dep.test.ts` — both PRs added a file with that name; #140's version is kept.

Those were the only two cherry-pick conflicts.

**Added** (`0d797e6`): #140 made the tongo import dynamic but left it as a `.then()/.catch()` chain, so it needed the same `try/catch` shape as the loaders you restructured. Your NOTE about webpack is right, and I measured it — packed install into a project with no optional peers, webpack 5 production build of an app importing `StarkZap` + `PrivySigner`:

| tongo loader | Result |
| --- | --- |
| `.then()/.catch()` | **1 error**, 7 warnings — `ERROR in .../confidential/tongo.js` |
| `try { await import() } catch` | **0 errors**, 8 warnings |

All eight remaining warnings are on lazy loaders, and the emitted bundle runs with both exports usable.

Also verified: typecheck, lint and prettier clean; 761 unit tests pass; `optional-peer-imports.test.ts` passes with both halves of the fix in place. Integration tests not run (they need a devnet).

Thanks for this one — the loader/`try`-block detail in particular was not obvious.

PR #141, opened 2026-08-19
fix: Make optional peer deps truly optional (supersedes #136)

micbakos:
Supersedes #136 by @l-henri, whose commit is carried here unchanged (authorship preserved). Moved into a branch on this repo because a fork PR cannot get green CI here: its workflow run sat at `action_required` and never produced a single check, and CI installs the privacy SDK from GitHub Packages with `secrets.GITHUB_TOKEN`, which a fork's read-only token cannot resolve against this org.

## Problem

starkzap declares its heavy dependencies as optional peers, but a consumer using none of those features still could not build. Two causes: peers imported statically in root-reachable modules, and lazy loaders written as `.then()/.catch()` chains, which webpack treats as a hard resolution failure rather than an optional import.

## Change

From #136, unchanged:

- **`ethers` out of the root graph.** Seven bridge modules move to `import type` and reach values through `await loadEthers(...)`, or a new synchronous `requireEthers(feature)` where the caller cannot await. `CCTPBridge`'s module-eval statics (`new Interface(...)`, checksummed addresses) become lazily cached getters.
- **All loaders restructured** to `try { await import(...) } catch` — avnu, ethers, `@solana/web3.js`, hyperlane. Error messages unchanged.
- **`tests/optional-peer-imports.test.ts`** walks the emitted module graph from `src/index.ts` and `src/cartridge.ts`, following static imports, re-exports and internal dynamic `import()`s, and asserts no optional peer is reachable statically.

Added here:

- **The tongo loader gets the same `try/catch` treatment.** #140 made it dynamic but left a `.then()/.catch()` chain.

## Dropped from #136, with reasons

#136's tongo work is superseded by #140, which landed after it was opened:

- **`src/confidential/tongoRuntime.ts` and `TongoConfidential.ready()`** — #140 removed the static tongo import by making `TongoConfidential.create(...)` an async factory. The module is loaded before an instance exists, so nothing needs a separate `ready()` step and the synchronous accessors never observe a half-built object. #136's approach reached the same goal but required `await ready()` before `address`, `recipientId`, `recipientFromAddress` and `getTongoAccount` would work.
- **`tests/tongo-optional-dep.test.ts`** — both PRs added a file with this name; #140's version is kept.

These were the only cherry-pick conflicts, and both were in tongo tests.

## Verified

Webpack 5 production build of an app importing `StarkZap` + `PrivySigner`, from a packed install into a project with **no** optional peers:

| tongo loader form | Result |
| --- | --- |
| `.then()/.catch()` (before) | **1 error**, 7 warnings — `ERROR in .../confidential/tongo.js` |
| `try { await import() } catch` (this branch) | **0 errors**, 8 warnings |

All eight warnings are on lazy loaders, which is the intended outcome: a missing peer becomes a runtime error only if the feature is used. The emitted bundle then runs and both exports are usable.

Also: typecheck, lint and prettier clean; 761 unit tests pass, 12 skipped; `optional-peer-imports.test.ts` passes with both halves of the fix present. Integration tests not run (they need a devnet).

Closes #136.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/bridge/ethereum/ContractRoutedEthereumBridge.ts
+++ b/src/bridge/ethereum/ContractRoutedEthereumBridge.ts
@@ -14,7 +14,8 @@ import type {
   EthereumTransactionDetails,
   EthereumWalletConfig,
 } from "@/bridge/ethereum/types";
-import { Contract, type ContractTransaction, type InterfaceAbi } from "ethers";
+import type { Contract, ContractTransaction, InterfaceAbi } from "ethers";
+import { requireEthers } from "@/connect/ethersRuntime";
 import { FeeErrorCause } from "@/types/errors";
 import type { WalletInterface } from "@/wallet";
 import CANONICAL_BRIDGE_ABI from "@/abi/ethereum/canonicalBridge.json";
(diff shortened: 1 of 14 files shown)


micbakos:
Closing in favour of #136, which now carries the same two commits after a rebase onto main. The work belongs on the original author's PR.

